Issue #12, opened 2022-12-22
Make cloudtemple_compute_network_adapter and cloudtemple_compute_virtual_disk importable


remilapeyre:
The API is missing the `networkId` and `datastoreClusterId` attribute so we cannot implement the import support in the provider yet.

PR #29, opened 2023-03-20
Make compute_network_adapter resource importable

aureldsn:
Partially fix issue #12


examples/resources/cloudtemple_compute_network_adapter/import.sh
```diff
@@ -0,0 +1 @@
+$ terraform import cloudtemple_compute_network_adapter.foo c74060bf-ebb3-455a-b0b0-d0dcb79f3d86
```
remilapeyre, line 1:
```suggestion
$ terraform import cloudtemple_compute_network_adapter.foo c74060bf-ebb3-455a-b0b0-d0dcb79f3d86

```

examples/resources/cloudtemple_compute_virtual_machine/import.sh
```diff
@@ -0,0 +1 @@
+$ terraform import cloudtemple_compute_virtual_machine.foo de2b8b80-8b90-414a-bc33-e12f61a4c05c
```
remilapeyre, line 1:
```suggestion
$ terraform import cloudtemple_compute_virtual_machine.foo de2b8b80-8b90-414a-bc33-e12f61a4c05c

```

remilapeyre (changes requested):
There can only be one network adapter per network and the ID of the network adapter is always the same than the one of the network?

aureldsn:
> There can only be one network adapter per network and the ID of the network adapter is always the same than the one of the network?

A virtual machine can have multiple network adapters with the same network ID

> Oh ok, I think I got it, the addition of the `network_id` attribute is unrelated to making `compute_network_adapter` importable, that is correct?

No, if I didn't made this change, after the `cloudtemple_compute_network_adapter` resource is imported when we run a tf plan, it indicates to add the `network_id` parameter instead of doing nothing

Issue #24, opened 2023-03-06
cloudtemple_compute_virtual_machine: Update guest_operating_system_moref value


aureldsn:
Currently, we cannot update the guest os value on a virtual machine without destroying it.

An API call is available to update this value: 

PATCH `/v1/vcenters/virtual_machines/{id}/guest`

```
{"guestOperatingSystemMoref":"debian10_64Guest"}
```

Moreover, there is a conflict between `guest_operating_system_moref` and `clone_virtual_machine_id`/`content_library_item_id`, this must be removed and changed by after the virtual machine creation from content library or clone, request an update of the guest os value (PATCH `/v1/vcenters/virtual_machines/{id}/guest`).

### Terraform Version

```
Terraform v1.3.9
on linux_amd64
+ provider registry.terraform.io/cloud-temple/cloudtemple v0.4.2
```

### Affected Resource(s)

- cloudtemple_compute_virtual_machine

### Terraform Configuration Files

```hcl
data "cloudtemple_compute_virtual_datacenter" "lab" {
  name = "DC-ITX7"
}

data "cloudtemple_compute_host_cluster" "lab" {
  name = "clu003-ucs02_LAB01"
}

data "cloudtemple_compute_datastore_cluster" "lab" {
  name = "sdrs003-LIVE_LAB01"
}

data "cloudtemple_compute_network" "lab" {
  name = "ISC-LAB-DC-ITX7"
}

resource "cloudtemple_compute_virtual_machine" "ads-test-002" {
  name = "ads-test-002"

  memory                 = 1 * 1024 * 1024 * 1024
  cpu                    = 1
  num_cores_per_socket   = 1
  cpu_hot_add_enabled    = true
  cpu_hot_remove_enabled = true
  memory_hot_add_enabled = true

  virtual_datacenter_id        = data.cloudtemple_compute_virtual_datacenter.lab.id
  host_cluster_id              = data.cloudtemple_compute_host_cluster.lab.id
  datastore_cluster_id         = data.cloudtemple_compute_datastore_cluster.lab.id
  guest_operating_system_moref = "debian10_64Guest"

  power_state = "off"

  tags = {
    created_by = "Terraform",
    env        = "lab",
  }

  lifecycle {
    ignore_changes = [virtual_datacenter_id]
  }
}
```

### Expected Behavior

When I change the value of `guest_operating_system_moref`, terraform should only update this value without any trouble.

### Actual Behavior

[...]

Issue #23, opened 2023-03-06
cloudtemple_compute_virtual_machine: Manage hardware_version value


aureldsn:
Can you add the possibility to upgrade the hardware_version value on a virtual machine?

The API call to get the available values for `hardware_version` is: `/v1/vcenters/virtual_machines/{id}/versions`

PR #31, opened 2023-03-21
Allow to update guest_operating_system_moref paramater

aureldsn:
Fix issue #24

### DIFF
--- a/docs/resources/compute_virtual_machine.md
+++ b/docs/resources/compute_virtual_machine.md
@@ -128,16 +128,16 @@ resource "cloudtemple_compute_virtual_machine" "content-library" {
 
 ### Optional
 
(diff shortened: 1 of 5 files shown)


remilapeyre:
Hi @aureldsn `guest_operating_system_moref` does not conflict with `content_library_item_id`? A content library item does not come with an operating system set?

aureldsn:
Hi @remilapeyre,
There is no conflict between `guest_operating_system_moref` and `content_library_item_id`.  
`content_library_item_id` is related to an OVF template or a vm template, so yes an operating system is well set but the API doesn't expose any info about it.
With my changes, if `guest_operating_system_moref` is set, it's only used after the virtual machine is built. This is very useful when we create a virtual machine with an old content library item and need to change the guest os with a value supported by the esx. We have the same issue with the virtual machine version (#23).

remilapeyre (approved):
This makes sense thanks!

Issue #13, opened 2022-12-22
Find how to implement deleting a running VM


remilapeyre:

aureldsn:
After internal discussions, the best user experience is to integrate the shutdown of the virtual machine before deleting it like other providers ([aws](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/ec2_instance.go#L2795), [vsphere](https://github.com/hashicorp/terraform-provider-vsphere/blob/main/vsphere/resource_vsphere_virtual_machine.go#L873-L882))

To avoid accidental destruction, we need to define the `prevent_destroy` variable on critical resources.

PR #33, opened 2023-03-22
Add a step to stop a running vm before deleting it

aureldsn:
Fix issue #13


remilapeyre:
Do you want to make this the default behavior or do you want to add an argument to control the behavior, a bit like `deletion_protection` exists in for AWS databases : https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance#deletion_protection ?

remilapeyre (approved):

PR #127, opened 2024-01-25
Updated datastore 'maintenance_mode' property type from string to boolean

pbesret:
**What changed :**
* Updated datastore 'maintenance_mode' property type from string to boolean


pbesret:
Note : Please wait for Compute API 1.87 to be release in production before merging.

Issue #238, opened 2026-05-18
VPC provider surface: datasources, resources, OpenIaaS association


pbesret:
**Is your feature request related to a problem? Please describe.**
As a user, I would like to be able to manage resources and datasources from the Cloud Temple VPC product.

**Describe the solution you'd like**
Resources and datasources from the Cloud Temple VPC, including the ability to associate a VPC network to an Open IaaS or VMware virtual machine.

**Additional context**
This feature is mandatory to implement the future VM instances product.

---

## Status — updated 2026-06-27

The VPC surface (delivered in v1.8.0, then removed in #362 and the SDK quarantined) has been **rebuilt on `rc/v1.9.0` in reviewable chunks**. **C0–C5 are merged**: the read surface (8 datasources) plus three resources — `cloudtemple_vpc_static_ip`, `cloudtemple_vpc_floating_ip`, `cloudtemple_vpc_floating_ip_binding` — are code-complete, each validated end-to-end on a live tenant before merge.

**Remaining for v1.9.0:** C6 (cross-surface live validation) and release plumbing (create the `v1.9.0` milestone, `rc/v1.9.0` → `main` + tag). A scope decision is pending: ship v1.9.0 without the internet gateway (recommended) or wait for it.

The VPC platform/API blockers found while building the surface are now **resolved platform-side** (latest deployment) or **deferred** — see "Platform/API items" below.

## v1.9.0 rebuild tracker (`rc/v1.9.0`)

Each chunk is cut from `rc/v1.9.0`, behind the standard review gates (PLAN → pre-commit → committed-diff) and an explicit human merge GO.

| Chunk | Scope | Issue | Status |
|---|---|---|---|
| C0 | Architecture plan + pre-code review gate | — | ✅ Done |
| C1 | Async client: `vpc_static_ip` CREATE is async, not sync | #348 | ✅ Merged (PR #364) |
| C2 | 8 read-only datasources + flatten helpers + golden gate | #366 | ✅ Merged (PR #365, merge `d1c68a0`) |
| C3 | Resource `cloudtemple_vpc_static_ip` (async create, never-drop, `source=custom` guard) | #367 | ✅ Merged (PR #368, merge `9d6d516`) |
| C4 | Floating-IP: client + resource `cloudtemple_vpc_floating_ip` | #370 (← #310 split) | ✅ Merged (PR #369, merge `b38aa84`) |
[...]

PR #239, opened 2026-05-18
feat(#238): Implemented VPC datasources

pbesret:


chrlesur:
Caduc / superseded. The VPC datasources, clients and helpers in this draft are fully delivered fresh on the `rc/v1.8.0` train by #301 (Phase 1 datasources) and #304 (Phase 2 reads) — re-implemented against the canonical Shiva VPC API (`/vpc/v1/...`), with the 403-as-not-found convention, the OpenIaaS VM `addresses` decode fix and the network-adapter VPC/IP reads, and validated live. This PR targets `dev` and is 133 commits behind rc. The only element not yet on rc is the `vpc_floating_ip` **write** resource, which is kept as a reference for VPC Phase 3 (to be re-implemented against the current rc VPC clients). VPC write resources + the VPC↔VM association are tracked under #238 (Phases 3-4). Closing as superseded.

PR #249, opened 2026-06-11
Client: retry transient errors (429/5xx/transport) while polling activity status

chrlesur:
Refs #245

Lot B of the non-complacent test plan (#264), built on top of the original draft.

## Problem

A single transient error (429, 5xx or a transport failure) while reading an activity status failed the whole operation — which kept running platform-side, leaving the resource orphaned outside the Terraform state.

## Changes

1. **Transient classification** (`isTransientAPIError`): 429, 5xx and transport-level failures are retryable during activity polling; everything else (4xx, decode errors) stays permanent. The budget is **bounded and consecutive** (`maxActivityReadRetries = 8`), reset by any successful read.
2. **Injectable polling loop** (`waitForActivityCompletion(ctx, id, readFn, backoff, options)`): the read function and the backoff are parameters, so the loop is unit tested without HTTP calls or sleeps.
3. **Pinned invariants** (8 test groups, no sleep): transient errors survived; budget bounded and consecutive; permanent errors never retried; a terminal `failed` activity never retried (retrying would hide real platform failures); initial not-found tolerated once, repeated not-found permanent; polling bounded by the injected backoff/context; context cancellation stops polling immediately.

## Validation

- `go build`/`go vet` OK; full unit suites green (`internal/client`, `internal/provider`, `internal/provider/helpers`).
- Adversarial review (`codex review`, default high config) before commit: refactor preserves the existing polling behavior, no actionable regression.
- The CI `unit` job runs all of this on the PR.


pbesret (approved):

chrlesur:
Pre-commit adversarial review (`codex review`, default high-reasoning config; reviewer model: Codex, pilot model: Claude): the refactor preserves the existing activity polling behavior and the new unit tests pass — no actionable regression.

VERDICT: VALIDATED — pending CI green on this PR, then merge into `rc/v1.8.0`.

chrlesur:

> The new transient retry loop keeps using the total attempt counter for the one-time not-found tolerance, which breaks the combined transient-plus-eventual-consistency case. The focused Go tests pass locally with a repository-local GOCACHE, but this edge case is not covered.$'\n'> $'\n'> Review comment:$'\n'> $'\n'> - [P2] Preserve first not-found tolerance after transient reads — /Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/activity.go:205-205$'\n'>   If a transient read failure happens before the activity becomes visible, `count` is already greater than 1 when the first successful read returns `nil`, so this existing eventual-consistency path treats the first not-found as a repeated not-found and aborts. This can still fail creates/updates during activity propagation when a 429/5xx/transport blip precedes the initial 404/nil; the not-found tolerance should be tracked independently from total polling attempts or only count successful activity reads.$'\n'> The new transient retry loop keeps using the total attempt counter for the one-time not-found tolerance, which breaks the combined transient-plus-eventual-consistency case. The focused Go tests pass locally with a repository-local GOCACHE, but this edge case is not covered.$'\n'> $'\n'> Review comment:$'\n'> $'\n'> - [P2] Preserve first not-found tolerance after transient reads — /Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/activity.go:205-205$'\n'>   If a transient read failure happens before the activity becomes visible, `count` is already greater than 1 when the first successful read returns `nil`, so this existing eventual-consistency path treats the first not-found as a repeated not-found and aborts. This can still fail creates/updates during activity propagation when a 429/5xx/transport blip precedes the initial 404/nil; the not-found tolerance should be tracked independently from total polling attempts or only count successful activity reads.

chrlesur:

> Christophe, local review only: GitHub publishing was blocked by the read-only/restricted environment. Tests also could not run because Go could not create its build dir under `/var/folders/...`.
> 
> 1. `url.Error` classification is broad. [api.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/api.go:385) treats any `*url.Error` as transient. Permanent examples include unsupported/misconfigured scheme, bad host from config, TLS trust/hostname/cert rejection, malformed proxy/proxy CONNECT auth failures, and some DNS/proxy configuration errors. In the polling loop this can delay surfacing the real config/transport error for 8 retryable failures, then the 9th consecutive failure becomes permanent. With Fibonacci backoff starting at 1s, that is roughly 54s plus request latency, not unbounded.
> 
> 2. Flapping endpoint: confirmed. [activity.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/activity.go:214) increments the consecutive failure budget only on transient read errors, and [activity.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/activity.go:225) resets it after any successful read. `WaiterOptions` is only logging, no timeout. I found no provider `Timeouts:` / `schema.ResourceTimeout`. So the overall bound is the caller `ctx` only. Operational worst case during apply: a resource operation can poll indefinitely, capped per interval at 30s backoff, until Terraform/user/orchestrator cancellation or terminal activity state.
> 
> 3. Not-found / 403 path: confirmed safe. `Activity.Read` calls `requireNotFoundOrOK(resp, 404)` at [activity.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/activity.go:96). 404 maps to `(false, nil)`, so first nil activity is retried once and repeated nil is permanent. 403 returns `fmt.Errorf("Access denied...")` at [api.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/api.go:365), not `StatusError`; `isTransientAPIError` returns false. Even if represented as `StatusError{Code:403}`, [api.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/api.go:388) returns false.
> 
> 4. Terminal failed activity: internally non-retryable, confirmed at [activity.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/activity.go:247). Small correction to the premise: I found two provider-level narrow rerun sites, not only one. `runVIFUpdateWithRetry` is bounded to 3 total attempts and only retries `IsTransientActivityFailure` ([openiaas_vif_retry.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/openiaas_vif_retry.go:56)); OpenIaaS network adapter create also retries up to 3 total attempts with the same narrow matcher ([resource_compute_iaas_opensource_network_adapter.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_network_adapter.go:97)). I did not find a broad caller retry that reruns arbitrary failed activities.
> 
> VERDICT: VALIDATED

Issue #264, opened 2026-06-12
CI: establish a platform-independent unit-test and static-analysis safety net


chrlesur:
## Summary

The provider has broad **surface** test coverage but almost no test layer that can run **without a live Cloud Temple platform**, and the CI does not execute any test on pull requests or pushes. This issue proposes establishing a platform-independent test and static-analysis safety net.

This is complementary to #8 (running the full integration suite in CI). #8 targets the acceptance/integration tests that require live credentials; this issue targets the missing unit-test layer and the CI lane that can run on every PR with no platform.

## Current state (observed on the repository)

- **195 test functions** total: 115 in `internal/provider`, 80 in `internal/client`.
- **Surface coverage is good**: 13/13 resources and 88/90 datasources have a test file under `internal/provider/tests/`.
- **Nature of the tests is the problem**: 101 of 103 provider test files are acceptance tests (`resource.Test` + `PreCheck`) and require `TF_ACC=1` plus a live platform. `.env.test.dist` lists ~40 environment variables (tenant, datacenter, vCenter, VM, buckets, …) that must point to existing objects. The `internal/client` tests are live-API integration tests as well.
- **Only one pure unit-test file exists**: `internal/provider/resource_compute_iaas_opensource_virtual_machine_unit_test.go` (3 tests), added with the `tx_checksumming` fix.
- **CI runs no tests**: in `.github/workflows/test.yml` only the `build` and `generate` (doc) jobs are active; the acceptance test matrix job and the scheduled drift run are commented out. `release.yml` runs no tests. No coverage is measured anywhere.

Net effect: on a multi-client provider, no automated check validates behavior on a PR, and the only way to exercise the suite is a manual `make testacc` / `make testclient` by an operator who holds platform access.

## Proposed scope

1. **Add a platform-independent CI lane** (on `pull_request` and `push`) that runs without any live credentials:
   - `go build ./...`
   - `go vet ./...`
   - `gofmt`/`terraform fmt` check
   - the pure unit tests (`go test` on the non-`TF_ACC` paths)
2. **Establish a unit-test layer for pure logic** that does not need the platform, starting with the highest state-safety value:
   - schema ↔ API mapping (flatten/expand)
   - diff/patch decisions and `optional`/`computed` field gating (the class of bug fixed in #247)
3. **Add coverage measurement** (`go test -cover`, surfaced in CI) to make the unit-test layer observable over time.
[...]

PR #265, opened 2026-06-12
OpenIaaS: bounded retry of VIF operations failing for transient platform reasons

chrlesur:
Refs #251

Supersedes #252. Final piece of the #264 state-safety plan, rebased on the post-Lot-A/B shape and hardened per the adversarial review verdicts.

## Design

1. **Narrow transient matcher** (`IsTransientActivityFailure`): only the known transient platform reason ("None of the workers were able to respond") is retryable — pinned by tests against arbitrary failure reasons (a MAC conflict must stay immediately fatal).
2. **`runVIFUpdateWithRetry`** — the critical invariant: **the payload is rebuilt against a freshly read live adapter before every attempt** (via the gated builders from #268). If a transiently-failed attempt actually converged server-side, the rebuilt payload is nil and the loop exits successfully instead of re-sending a stale PATCH the platform would reject as a VPC static-IP self-conflict. Wired into both the VM path (`osNetworkAdapterUpdate`) and the standalone adapter update.
3. **Bounded everywhere**: 3 TOTAL attempts; no call after success; permanent failures and missing adapters fail immediately; the injectable API surface (read/update/wait/sleep/classifier) makes the six invariants unit-testable without HTTP or sleeps.
4. **Create path** (standalone adapter): bounded retry with the anti-duplicate guard — only the adapter referenced by the FAILED activity itself (`ConcernedItems`) is cleaned up before retrying, never a list-diff.

## Validation

- 6 `TestVIFRetry*` invariants + `TestIsTransientActivityFailure` matcher pin — all green; full unit suites green; `go build`/`go vet` OK.
- Adversarial review (`codex review`, default high config) before commit: no actionable regression.
- The CI `unit` job runs all of this on the PR.


chrlesur:
Pre-commit adversarial review (`codex review`, default high-reasoning config; reviewer model: Codex, pilot model: Claude): no actionable regression. The implementation incorporates the two prior review verdicts on this topic: the stale-payload re-PATCH risk is eliminated by the live-state recompute before every attempt, and the transient matcher stays narrow (pinned by tests).

chrlesur:

> The update retry path is covered by tests and the targeted packages pass locally with an in-workspace Go cache, but the create retry path can still duplicate VIFs if the cleanup attempt is not confirmed. That is a blocking state/platform safety issue for this change.$'\n'> $'\n'> Review comment:$'\n'> $'\n'> - [P1] Abort retry when VIF cleanup is unconfirmed — /Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_network_adapter.go:122-123$'\n'>   When a transient create activity reports a half-created adapter but `Delete` fails or the delete activity itself fails, this code ignores the cleanup failure and continues to the next create attempt. In that scenario the original VIF can remain platform-side and the retry may create a duplicate adapter, which defeats the anti-duplicate guard introduced here; the retry should stop or surface the cleanup failure unless deletion is confirmed.$'\n'> The update retry path is covered by tests and the targeted packages pass locally with an in-workspace Go cache, but the create retry path can still duplicate VIFs if the cleanup attempt is not confirmed. That is a blocking state/platform safety issue for this change.$'\n'> $'\n'> Review comment:$'\n'> $'\n'> - [P1] Abort retry when VIF cleanup is unconfirmed — /Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_network_adapter.go:122-123$'\n'>   When a transient create activity reports a half-created adapter but `Delete` fails or the delete activity itself fails, this code ignores the cleanup failure and continues to the next create attempt. In that scenario the original VIF can remain platform-side and the retry may create a duplicate adapter, which defeats the anti-duplicate guard introduced here; the retry should stop or surface the cleanup failure unless deletion is confirmed.

chrlesur:

> **Findings**
> 
> 1. Blocking: the VM update path can leave a VM powered off after VIF retry failure. `handleUpdateOSDevices` powers off when needed, then returns immediately on any disk/VIF update diagnostic before reaching the power-on block. A VIF transient storm is bounded to 3 attempts, but after the third failure `osNetworkAdapterUpdate` returns an error and the VM stays off until external recovery. See [resource_compute_iaas_opensource_virtual_machine.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1273), [resource_compute_iaas_opensource_virtual_machine.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1306), [resource_compute_iaas_opensource_virtual_machine.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1311).
> 
> 2. Blocking: create cleanup trusts `ConcernedItems` too much before deletion. The filter is exact (`item.Type == "network_adapter"`), but ownership is not proven: no validation that the adapter belongs to this create operation, or even to `vmID`, before `Delete`. If the failed activity ever carries a stale/wrong/shared concerned adapter, this can delete a legitimate VIF. See [resource_compute_iaas_opensource_network_adapter.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_network_adapter.go:119).
> 
> 3. Risk: `IsTransientActivityFailure` is substring-based on `state.Reason`. A permanent failure embedding `None of the workers were able to respond` will be retried and, on create, trigger cleanup. The 3-total-attempt bound is enforced in both update wrapper and create loop, but the classifier is still spoofable by message text. See [activity.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/activity.go:17), [activity.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/activity.go:28), [openiaas_vif_retry.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/openiaas_vif_retry.go:65), [resource_compute_iaas_opensource_network_adapter.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_network_adapter.go:97).
> 
> On attack 1 specifically: inside `runVIFUpdateWithRetry`, `actual == nil` fails closed with `network adapter ... not found`; it does not PATCH and does not mutate Terraform state. But the underlying client really does map 403 to nil via `requireNotFoundOrOK(_, 403)`, so the error is misleading, and surrounding refresh paths can still drop state on nil outside the retry wrapper. See [compute_iaas_opensource_network_adapter.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/compute_iaas_opensource_network_adapter.go:45), [api.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/api.go:359), [openiaas_vif_retry.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/openiaas_vif_retry.go:70).
> 
> On cancellation: no retry continues after cancellation during sleep, but both create/update return the previous transient error rather than `ctx.Err()`, which hurts diagnosis. I could not publish a GitHub review (`api.github.com` unreachable) or run tests (`go test` cannot create its build dir in the read-only sandbox).
> 
> VERDICT: NEEDS CHANGES - VM VIF failures can leave a powered-off VM, create cleanup can delete an unproven ConcernedItems adapter, and transient matching is substring-spoofable.

chrlesur:
The P1/P2 findings of the re-review campaign on the create path are resolved by #276, merged on explicit human GO after the full strict review cycle (3 attack iterations + 2 committed-diff passes, verdicts quoted verbatim on the PR).

Issue #267, opened 2026-06-12
OpenIaaS: gate the VM-properties and VIF PATCH payloads on real live divergence


chrlesur:
State-safety P0 batch from the adversarial test plan (see #264, plan comment).

## Problem

`openIaasVirtualMachineUpdate` sends an **unconditional** full-properties PATCH (no `HasChange` guard, no live comparison): `Name`, `CPU`, `NumCoresPerSocket`, `Memory`, `BootFirmware`, `SecureBoot`, `AutoPowerOn`, `HighAvailability` are pushed from `d.Get` on every update — including the create→update chaining. This is the same class as the #246 VIF incident:

- values inherited through Computed are re-pushed as write intent;
- `SecureBoot bool` (no `omitempty`) always serializes — an absent Optional+Computed boolean is sent as `false` and can mutate a live setting;
- an explicit `secure_boot = false` is swallowed by the post-create state merge on first apply (same mechanism as the fixed `tx_checksumming`).

## Changes

1. Extract `buildOpenIaasVMPropertiesPatch(live, desired) -> (req, changed, needsReboot)`: pure, unit-testable; each field included only on real divergence from the live API state; `secure_boot` gated on the raw configuration (`*bool`); reboot requirement derived from the real cpu/memory/cores divergence instead of `HasChange`.
2. `UpdateOpenIaasVirtualMachineRequest.SecureBoot/AutoPowerOn` become `*bool,omitempty` (an absent boolean is omitted from the PATCH, an explicit `false` is expressible).
3. Extract `buildOpenIaasVIFPatch(desired, actual, txWant) -> *UpdateOpenIaasNetworkAdapterRequest` (nil when converged) from `osNetworkAdapterUpdate` — payload inspectable by tests without client tricks.
4. Registry test `TestOptionalComputedBooleansHavePatchGuards`: every Optional+Computed boolean reaching an API request must be raw-config-gated or documented read-only (deliberately annoying allowlist).
5. Adversarial unit tests for all of the above (no-PATCH-when-converged, explicit-false-on-first-apply, absent-never-pushed).

No schema change, no state encoding change. Refs the non-complacent test plan in #264.

chrlesur:
[...]

PR #268, opened 2026-06-12
OpenIaaS: gate the VM-properties and VIF PATCH payloads on real live divergence

chrlesur:
Refs #267

State-safety P0 batch (Lot A) of the non-complacent test plan (#264).

## Changes

1. **`buildOpenIaasVMPropertiesPatch(live, desired)` → `(req, changed, needsReboot)`** — pure and unit-testable. Each property is included in the PATCH only when it actually diverges from the live API state; the unconditional full-properties PATCH (same class as the #246 VIF incident) is gone. The reboot requirement now derives from the real cpu/memory/cores divergence instead of `HasChange`.
2. **`secure_boot` raw-config gating** — Optional+Computed: the merged plan value is not write intent; only an explicit configuration can produce a write (pattern: `tx_checksumming`). `SecureBoot`/`AutoPowerOn` become `*bool,omitempty` in `UpdateOpenIaasVirtualMachineRequest` (absent = omitted from the payload, explicit `false` expressible).
3. **`buildOpenIaasVIFPatch(desired, actual, txWant)`** — extracts the VIF payload construction (nil when converged) from `osNetworkAdapterUpdate`; payloads are now inspected directly by tests.
4. **Registry test `TestOptionalComputedBooleansHavePatchGuards`** — walks every resource schema and fails when an Optional+Computed boolean is not in the documented guard registry (5 gated, 5 documented UNGATED-LEGACY pending the VMware parity batch, Lot D).

## Behavior notes

- A converged VM after a marketplace deploy now produces **zero** properties PATCH during the create→update chaining.
- `auto_power_on` (plain Optional) keeps its Terraform contract (absent = false) but is only sent when diverging from live — previous behavior always serialized it.
- No schema change, no state encoding change.

## Validation

- 17 new test cases (`TestBuildOpenIaasVMPropertiesPatch`, `TestBuildOpenIaasVIFPatch`, registry) + existing suites green; `go build`/`go vet` OK.
- Adversarial review (`codex review`, default gpt-5.5 high config) on the uncommitted diff before commit: no actionable bug (the `internal/*/tests` build failure is pre-existing and outside this diff).
- The new CI `unit` job runs all of this on the PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ BUG FIXES :
   * Fixed a bug causing an explicit `tx_checksumming = false` to never be sent to the API.
   * Fixed a bug causing `tx_checksumming` to be pushed from resource `cloudtemple_compute_iaas_opensource_virtual_machine` when it was not explicitly configured.
   * Fixed a bug causing the cloud-init config drive (`XO CloudConfigDrive`) to be captured as a managed `os_disk` at creation on resource `cloudtemple_compute_iaas_opensource_virtual_machine`, producing a permanent removal drift.
+  * Fixed a bug causing resource `cloudtemple_compute_iaas_opensource_virtual_machine` to send an unconditional full-properties update on every apply. Properties are now only patched when they actually diverge from the live API state, and `secure_boot` is only sent when explicitly configured.
 
 IMPROVEMENTS :
 
--- a/internal/client/compute_iaas_opensource_virtual_machine.go
+++ b/internal/client/compute_iaas_opensource_virtual_machine.go
@@ -128,13 +128,17 @@ func (v *OpenIaaSVirtualMachineClient) Read(ctx context.Context, id string) (*Op
 }
 
 type UpdateOpenIaasVirtualMachineRequest struct {
+	// All fields are optional (PATCH semantics): callers only set the
+	// fields that actually diverge from the live state. The booleans are
+	// pointers so an absent value is omitted from the payload while an
+	// explicit false stays expressible (#267).
 	Name              string `json:"name,omitempty"`
 	CPU               int    `json:"cpu,omitempty"`
 	NumCoresPerSocket int    `json:"numCoresPerSocket,omitempty"`
 	Memory            int    `json:"memory,omitempty"`
-	SecureBoot        bool   `json:"secureBoot"`
+	SecureBoot        *bool  `json:"secureBoot,omitempty"`
 	BootFirmware      string `json:"bootFirmware,omitempty"`
-	AutoPowerOn       bool   `json:"autoPowerOn"`
+	AutoPowerOn       *bool  `json:"autoPowerOn,omitempty"`
 	HighAvailability  string `json:"highAvailability,omitempty"`
 }
 
--- a/internal/provider/optional_computed_booleans_test.go
+++ b/internal/provider/optional_computed_booleans_test.go
@@ -1,0 +1,73 @@
+package provider
+
+import (
+	"fmt"
+	"sort"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// optionalComputedBooleanGuards is the registry of every Optional+Computed
+// boolean attribute of the provider resources. This list is DELIBERATELY
+// annoying to maintain: an Optional+Computed boolean read from d.Get cannot
+// distinguish an explicit user choice from a value inherited from the live
+// API, so pushing it into a PATCH without an explicit raw-config guard
+// silently mutates production resources (#246 class, #267).
+//
+// Before adding an entry here, either:
+//   - gate the attribute on d.GetRawConfig() evidence before any API write
+//     (pattern: osAdapterTxConfigured / openIaasVMDesiredPropertiesFromResourceData), or
+//   - document below why the attribute never reaches an API request.
+var optionalComputedBooleanGuards = map[string]string{
+	"cloudtemple_compute_iaas_opensource_virtual_machine.secure_boot":                        "raw-config gated in openIaasVMDesiredPropertiesFromResourceData (#267)",
+	"cloudtemple_compute_iaas_opensource_virtual_machine.os_disk.connected":                  "UNGATED-LEGACY: reconciled against live state in handleUpdateOSDevices; connect/disconnect paths use live VBD evidence",
+	"cloudtemple_compute_iaas_opensource_virtual_machine.os_network_adapter.attached":        "deprecated attribute, never written from the VM path",
+	"cloudtemple_compute_iaas_opensource_virtual_machine.os_network_adapter.tx_checksumming": "raw-config gated through osAdapterTxConfigured (map[string]*bool)",
+	"cloudtemple_compute_iaas_opensource_network_adapter.tx_checksumming":                    "raw-config gated in openIaasNetworkAdapterUpdate (txConfigured)",
+	"cloudtemple_compute_virtual_machine.boot_options.enter_bios_setup":                      "UNGATED-LEGACY: VMware parity batch (Lot D of the #264 plan)",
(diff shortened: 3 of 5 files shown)


chrlesur:
Pre-commit adversarial review (`codex review`, default high-reasoning config; reviewer model: Codex, pilot model: Claude): no actionable bug found in the diff — targeted tests for the touched provider/client/helper packages pass; the `internal/*/tests` build failure is pre-existing and outside this diff.

VERDICT: VALIDATED — pending CI green on this PR, then merge into `rc/v1.8.0`.

chrlesur:

> No actionable regression introduced by the modified diff was identified. The targeted Go tests for the modified provider/client packages pass with a sandbox-local Go cache.$'\n'> No actionable regression introduced by the modified diff was identified. The targeted Go tests for the modified provider/client packages pass with a sandbox-local Go cache.

chrlesur:

> **Findings**
> 
> 1. Blocking: `Optional+Computed` inherited values can still reach the PATCH for `num_cores_per_socket` and `boot_firmware`.
> 
> `secure_boot` is correctly raw-config gated: absent/inherited Computed values become `nil` and cannot patch ([resource_compute_iaas_opensource_virtual_machine.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1053), [line 1101](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1101)).
> 
> But `num_cores_per_socket` and `boot_firmware` are `Optional+Computed` and are read from `d.Get` without raw-config evidence ([line 1046](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1046), [line 1048](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1048)). The builder only protects zero/empty values ([line 1083](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1083), [line 1093](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1093)). A stale inherited non-zero/non-empty state value can therefore be pushed if live differs, especially with stale state or `-refresh=false`. That violates invariant 1 as stated.
> 
> Field walk:
> - `name`: Required, not Computed. Empty string is guarded out; likely OK, but schema does not reject empty names.
> - `cpu`: Required, not Computed. Zero guard is domain-safe only if CPU 0 is invalid, but schema does not enforce `>=1`.
> - `memory`: Required, not Computed. Same issue: memory 0 is probably invalid, but accepted by schema and silently ignored on update.
> - `num_cores_per_socket`: Optional+Computed. Zero guard protects absent create/update, but inherited stale non-zero can patch.
> - `boot_firmware`: Optional+Computed. Empty guard protects absent, but inherited stale `bios`/`uefi` can patch.
> - `high_availability`: Optional with default `"disabled"`, not Computed. Default-as-desired is Terraform-managed behavior.
> - `secure_boot`: Correct raw-config gate.
> - `auto_power_on`: Optional, not Computed. Always treated as desired bool.
> 
> 2. `needsReboot` is mostly safer than old `HasChange`, but the zero guard creates an invalid-config hole.
> 
> Old rebooted and new does not when `HasChange(cpu|memory|num_cores_per_socket)` was true but live already matched desired: create->update chaining, marketplace/template already applied values, or external convergence before apply. That is good.
> 
> New reboots where old might not when `HasChange` is false but the live VM diverges from desired at update time. That is also safer for CPU/memory/cores, because the old code would still send a full PATCH without the reboot bracket.
> 
> Skipped required power-off risk: only if the intended desired value is zero, because the field is omitted and `needsReboot` remains false. CPU/memory/cores zero should likely be invalid, but the schema does not reject it. Also, boot firmware / secure boot power-off requirements are not modeled; this is not introduced by PR #268, but the current gating still assumes only CPU/memory/cores require reboot.
> 
> 3. `403 -> nil` read is fail-closed for the VM-properties PATCH, but “ALWAYS reads the VM first” is not strictly true.
> 
> `Read` calls `requireNotFoundOrOK(resp, 403)`, and because `notFoundCode` is `403`, the `case 404, notFoundCode` branch maps 403 to `found=false, err=nil` ([api.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/api.go:359), [VM Read](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/compute_iaas_opensource_virtual_machine.go:117)). The update then checks `vm == nil` and returns before power/PATCH ([line 773](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:773), [line 777](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:777)). So for Zone A PATCH gating: fail-closed.
> 
> Caveat: replication policy changes run before that VM read ([line 738](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:738)). So the whole update path does not literally read the VM first.
> 
> 4. `auto_power_on=false` after marketplace create is consistent with previous behavior, but not clearly visible at first plan.
> 
> Current code always sets desired `AutoPowerOn` from `d.Get("auto_power_on")` ([line 1051](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1051)) and patches if live differs ([line 1105](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1105)). Previous code also always sent `AutoPowerOn: d.Get(...)`, and the old client JSON field was non-omitempty, so absent config already forced `false`.
> 
> So this is not a regression. But on create->update chaining, the user will not see “platform set true, provider will push false” as a distinct drift in the initial plan, because the platform-side true appears only after deploy. If the intended contract is really “absent means false”, add `Default: false` for explicitness.
> 
> 5. Yes, error paths can leave a VM powered off.
> 
> For VM properties: after power-off, any error in `Update` or `WaitForCompletion` returns before the power-on bracket ([line 791](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:791), [line 807](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:807), [line 819](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:819)). Same pattern exists in `handleUpdateOSDevices`: disk/VIF update errors after power-off return before power-on ([line 1273](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1273), [line 1295](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1295), [line 1306](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1306), [line 1311](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1311)).
> 
> This is pre-existing, but it matters here because the new live-divergence logic can decide to enter that bracket.
> 
> Tests not run: sandbox is read-only, and Go/git tooling attempted writes under `/tmp`.
> 
> VERDICT: NEEDS CHANGES - blocking reasons: `num_cores_per_socket` and `boot_firmware` can still PATCH inherited Optional+Computed values; zero-value guards rely on unenforced schema invariants for CPU/memory/cores; power-off error paths can leave the VM off.

Issue #271, opened 2026-06-12
State-safety: VMware boot_options gating, secrets preserved on refresh, read errors are not deletions


chrlesur:
Lot D (final batch) of the non-complacent test plan (#264).

## Problems

1. **`cloudtemple_compute_virtual_machine` re-pushed the full `boot_options` block on every update** (the create chains into update): the block is Optional+Computed, so values merely inherited from the live state — including the three Optional+Computed booleans `boot_retry_enabled`, `enter_bios_setup`, `efi_secure_boot_enabled`, serialized without `omitempty` — were re-sent as write intent (#246 class).
2. **`cloudtemple_iam_personal_access_token` erased `secret_id` from the state on refresh**: the API only returns the secret at creation; the Read overwrote the only stored copy with an empty value.
3. **A transient API error during refresh dropped `cloudtemple_compute_iaas_opensource_replication_policy` and `cloudtemple_compute_iaas_opensource_virtual_disk` from the state** (`if obj == nil || err != nil { d.SetId(\"\") }`): the next apply would create a duplicate of a resource that still exists.

## Changes

- `buildVMwareBootOptionsFromRaw` (pure, unit-tested): the block is only sent when explicitly configured; its booleans become `*bool,omitempty` and are only carried when explicitly set in the raw configuration.
- The PAT Read only overwrites `secret_id` when the API actually returns a secret.
- Read errors now surface as errors; only a genuine not-found clears the id.
- Registry updated: 3 booleans move to gated; the 2 remaining `os_network_adapter` legacy entries get precise guard documentation (full gating tracked in the #264 plan).

Refs #264.

chrlesur:
Merged into `rc/v1.8.0` via #272. With this, the four batches of the #264 state-safety plan are delivered (A: OpenIaaS PATCH gating #268, B: activity retry invariants #249, C: datasource walker #270, D: this one). Stays open for the aggregated `Closes #271` in the `rc/v1.8.0 -> main` PR.

chrlesur:
FF-4 (#279) and FF-5 (#280, merged on explicit human GO) close the Zone C deep-review findings: every boot_options attribute is explicitly gated, and an OpenIaaS deletion is only accepted under strict, complete listing evidence. The residual nil-drop sweep outside OpenIaaS (VMware, object storage, PAT) is the next item of the remediation series.

PR #272, opened 2026-06-12
State-safety: VMware boot_options gating, secrets preserved on refresh, read errors are not deletions

chrlesur:
Refs #271

Lot D (final batch) of the non-complacent test plan (#264).

## Changes

| Fix | Class |
|---|---|
| `boot_options` only sent when explicitly configured; its three Optional+Computed booleans become `*bool,omitempty` and are only carried when explicitly set (`buildVMwareBootOptionsFromRaw`, pure and unit-tested) | Unrequested PATCH (#246 class) — the VMware create chains into update |
| `cloudtemple_iam_personal_access_token` Read no longer erases `secret_id` (API returns the secret only at creation) | Secret destroyed on refresh |
| `cloudtemple_compute_iaas_opensource_replication_policy` and `..._virtual_disk` Reads: a transient API error is an error, not a deletion — only a genuine not-found clears the id | State destruction → duplicate creation |
| Registry: 3 `boot_options` booleans now gated; the 2 remaining `os_network_adapter` legacy entries get precise guard documentation | Guard registry hygiene |

## Validation

- New `TestBuildVMwareBootOptionsFromRaw` (unconfigured booleans omitted even when merged true; explicit false carried); registry test green; full unit suites green; `go build`/`go vet` OK; docs unchanged (no schema change).
- Adversarial review (`codex review`, default high config) before commit — its first pass caught a compile break in `internal/client/tests` (BootOptions literal), fixed; second pass: no regression.

With this, the four batches (A: OpenIaaS PATCH gating, B: activity retry invariants, C: datasource walker, D: VMware parity + secrets + read errors) of the #264 state-safety plan are delivered. Remaining on the plan: the full `os_network_adapter` raw-config gating (documented legacy) and the #265 bounded VIF retry, rebased on this final shape.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ BUG FIXES :
   * Fixed a bug causing resource `cloudtemple_compute_iaas_opensource_virtual_machine` to send an unconditional full-properties update on every apply. Properties are now only patched when they actually diverge from the live API state, and `secure_boot` is only sent when explicitly configured.
   * Fixed a bug causing an in-progress operation to be reported as failed when a transient error (429, 5xx or a transport failure) occurred while polling its activity status. Transient read failures are now retried with a bounded consecutive budget.
   * Fixed datasources `cloudtemple_compute_iaas_opensource_pools`, `cloudtemple_compute_iaas_opensource_virtual_machine`, `cloudtemple_compute_iaas_opensource_virtual_machines`, `cloudtemple_compute_iaas_opensource_virtual_disk` and `cloudtemple_compute_iaas_opensource_virtual_disks` failing at read time with an `Invalid address to set` error: the schemas now declare every attribute emitted by the flatten helpers.
+  * Fixed resource `cloudtemple_compute_virtual_machine` re-sending the full `boot_options` block (including values merely inherited from the live state) on every update. The block is now only sent when explicitly configured, and its booleans only when explicitly set.
+  * Fixed resource `cloudtemple_iam_personal_access_token` erasing `secret_id` from the state on refresh (the API only returns the secret at creation).
+  * Fixed resources `cloudtemple_compute_iaas_opensource_replication_policy` and `cloudtemple_compute_iaas_opensource_virtual_disk` being dropped from the Terraform state when a transient API error occurred during refresh, which made the next apply create a duplicate.
 
 IMPROVEMENTS :
 
--- a/internal/client/compute_virtual_machine.go
+++ b/internal/client/compute_virtual_machine.go
@@ -93,12 +93,16 @@ type VirtualMachineBootOptions struct {
 }
 
 type BootOptions struct {
+	// The booleans are pointers so an absent value is omitted from the
+	// payload while an explicit false stays expressible: the matching
+	// schema attributes are Optional+Computed and a merged value is not
+	// write intent (#264 plan, Lot D).
 	BootDelay            int    `json:"bootDelay"`
 	BootRetryDelay       int    `json:"bootRetryDelay"`
-	BootRetryEnabled     bool   `json:"bootRetryEnabled"`
-	EnterBIOSSetup       bool   `json:"enterBIOSSetup"`
+	BootRetryEnabled     *bool  `json:"bootRetryEnabled,omitempty"`
+	EnterBIOSSetup       *bool  `json:"enterBIOSSetup,omitempty"`
 	Firmware             string `json:"firmware"`
-	EFISecureBootEnabled bool   `json:"efiSecureBootEnabled"`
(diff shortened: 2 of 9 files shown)


chrlesur:
Pre-commit adversarial review (`codex review`, default high-reasoning config; reviewer model: Codex, pilot model: Claude): the first pass caught a compile break introduced in `internal/client/tests` by the BootOptions pointer migration — fixed (`boolPtr` literals). Second pass: no regression introduced; the remaining `internal/client/tests` build failure (`PAT().List` signature) is pre-existing and unrelated.

chrlesur:

> I did not identify an introduced bug in the changed files. Targeted provider/client package tests compile for the directly affected packages; broader test failures are from unchanged PAT List call sites outside this diff.$'\n'> I did not identify an introduced bug in the changed files. Targeted provider/client package tests compile for the directly affected packages; broader test failures are from unchanged PAT List call sites outside this diff.

chrlesur:

> Findings:
> 
> 1. Blocking: partial `boot_options` can still push zero values.
> When raw `boot_options` is absent, `bootOptions` is omitted from the PATCH JSON, so it should not revert live boot settings by itself: [compute_virtual_machine.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/compute_virtual_machine.go:225), [resource_compute_virtual_machine.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_virtual_machine.go:1221).
> But when the block is present, `boot_delay`, `boot_retry_delay`, and `firmware` are always copied from merged `d.Get` into non-pointer JSON fields: [resource_compute_virtual_machine.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_virtual_machine.go:1199). On create, a partial block can therefore send `bootDelay:0`, `bootRetryDelay:0`, `firmware:""` while only the configured booleans are gated. That can reset live/template values or send an invalid/empty required nested field.
> 
> 2. Create-time behavior is mostly preserved, but inherits the partial-block bug.
> VMware create always chains into `updateVirtualMachine(..., customizing=true)`: [resource_compute_virtual_machine.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_virtual_machine.go:1087). If `boot_options` is explicitly configured, it is still applied during that post-create update. If absent, it is omitted. Marketplace/template users relying on explicit boot options are fine only when they set the full object safely; partial blocks remain unsafe.
> 
> 3. PAT secret guard looks one-way safe.
> `secret_id` is set on create, and refresh only overwrites it when `token.Secret != ""`: [resource_iam_personal_access_token.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_iam_personal_access_token.go:101), [resource_iam_personal_access_token.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_iam_personal_access_token.go:132). With the current client model, empty string cannot distinguish “field omitted by API” from “intentional empty secret”; since PAT has no update/rotation path, clearing state on empty would be more dangerous.
> 
> 4. Blocking: the 403-as-deletion path still exists in both resources.
> Both reads call client `Read`, then `SetId("")` on `nil`: [replication_policy.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_replication_policy.go:130), [virtual_disk.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_disk.go:186). The clients use `requireNotFoundOrOK(resp, 403)`, and because `403` is passed as the not-found code, `403` returns `(false, nil)`: [api.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/client/api.go:359). So access denial still silently removes state. Consistent with the #275 strict-evidence rule, these reads should fail closed or require independent strict confirmation before `SetId("")`.
> 
> 5. Registry entries are accurate.
> The two `UNGATED-LEGACY` entries describe the residual VMware `os_network_adapter` risk correctly: `auto_connect` is pushed in the general adapter PATCH when any field in that adapter changes, while `connected` connect/disconnect is per-attribute gated: [optional_computed_booleans_test.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/optional_computed_booleans_test.go:31), [resource_compute_virtual_machine.go](/Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_virtual_machine.go:1453).
> 
> VERDICT: NEEDS CHANGES with blocking reasons: partial VMware boot_options still sends zero/empty non-pointer Optional+Computed fields, and replication policy/virtual disk Read still treat 403 as deletion without strict independent confirmation.

Issue #273, opened 2026-06-12
OpenIaaS: clean the legacy config-drive pollution from os_disk on refresh, and stop crashing on deleted devices


chrlesur:
Remediation series on `rc/v1.8.0` (P1 of the #264 plan, follow-up of #255).

## Problems

1. **Legacy state pollution**: #256 prevents the cloud-init config drive (`XO CloudConfigDrive`) from being captured at creation, but states polluted by pre-1.8.0 creations keep the drive as a managed `os_disk`: the Read re-reads it by id and the permanent removal drift continues for those clients today.
2. **Crash on refresh**: when an `os_disk` or `os_network_adapter` referenced in the state no longer exists platform-side, the Read passes the nil API answer to the flatten helper and panics (nil dereference) — same class as the #234 crash.

## Changes

- On the VM Read, a state `os_disk` is dropped (with an explicit **warning diagnostic**) only under the strict live evidence already used at creation: exact `XO CloudConfigDrive` name AND read-only VBD on this VM (`IsPlatformManagedDisk`). Ambiguous disks (same name but writable, read-only elsewhere, missing VBD) stay in the state and require the documented manual remediation. No platform-side change is ever made — this is a state-shape correction only.
- A deleted device (nil API answer) is dropped from the state list instead of crashing: the refresh reflects reality and the next plan shows the divergence.
- The decision logic is extracted (`classifyOSDiskOnRead`) and pinned by adversarial unit tests (the P1 cases of the #264 plan).

Refs #255, refs #264.

chrlesur:
Merged into `rc/v1.8.0` via #274. The legacy config-drive pollution is cleaned at the first refresh (strict live evidence, explicit warning, no platform-side change) and the refresh no longer crashes on devices deleted platform-side. Stays open for the aggregated `Closes #273` in the `rc/v1.8.0 -> main` PR.

PR #274, opened 2026-06-12
OpenIaaS: clean the legacy config-drive pollution on refresh, stop crashing on deleted devices

chrlesur:
Refs #273, refs #255

Remediation series on `rc/v1.8.0` — P1 of the #264 plan (the only drift existing clients still suffer today).

## Changes

1. **Legacy config-drive cleanup on refresh** — `classifyOSDiskOnRead` drops a state `os_disk` entry only under the strict live evidence already used at creation (exact `XO CloudConfigDrive` name AND read-only VBD on this VM): the permanent removal drift stops for clients polluted by pre-1.8.0 creations. An explicit **warning diagnostic** surfaces every cleanup. Ambiguous disks (writable, read-only elsewhere, no VBD) stay managed. **No platform-side change is ever made** — state-shape correction only.
2. **No more crash on deleted devices** — an `os_disk` or `os_network_adapter` referenced in the state but deleted platform-side made the refresh panic on the nil API answer (#234 class); the stale entry is now dropped so the next plan reflects reality.

## Validation

- 7 adversarial cases for `classifyOSDiskOnRead` (the P1 matrix of the #264 plan: deleted, cleaned, and the four ambiguous keep-cases) + the existing 5-case `IsPlatformManagedDisk` matrix — all green; full unit suites green.
- Adversarial review (`codex review`, default high config) before commit: no actionable regression.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ BUG FIXES :
   * Fixed resource `cloudtemple_iam_personal_access_token` erasing `secret_id` from the state on refresh (the API only returns the secret at creation).
   * Fixed resource `cloudtemple_compute_virtual_machine` re-sending the full `boot_options` block (including values merely inherited from the live state) on every update. The block is now only sent when explicitly configured, and its booleans only when explicitly set.
   * Fixed datasources `cloudtemple_compute_iaas_opensource_pools`, `cloudtemple_compute_iaas_opensource_virtual_machine`, `cloudtemple_compute_iaas_opensource_virtual_machines`, `cloudtemple_compute_iaas_opensource_virtual_disk` and `cloudtemple_compute_iaas_opensource_virtual_disks` failing at read time with an `Invalid address to set` error: the schemas now declare every attribute emitted by the flatten helpers.
+  * Fixed Terraform states polluted by pre-1.8.0 creations that captured the cloud-init config drive (`XO CloudConfigDrive`) as a managed `os_disk`: the refresh now drops the drive from the state — with an explicit warning and only under strict live evidence (exact XO name and read-only attachment to this virtual machine) — which stops the permanent removal drift for existing clients. Ambiguous disks stay in the state. No platform-side change is ever made.
+  * Fixed a crash (nil dereference) during refresh when an `os_disk` or `os_network_adapter` referenced in the state no longer exists platform-side: the deleted device is now dropped from the state so the next plan reflects reality.
 
 IMPROVEMENTS :
 
--- a/internal/provider/helpers/helper_compute_iaas_opensource_virtual_machine.go
+++ b/internal/provider/helpers/helper_compute_iaas_opensource_virtual_machine.go
@@ -72,7 +72,7 @@ func FlattenOpenIaaSVirtualMachine(vm *client.OpenIaaSVirtualMachine) map[string
 // it attaches at deploy time.
 const cloudConfigDriveName = "XO CloudConfigDrive"
 
-// isPlatformManagedDisk reports whether the disk is managed by the platform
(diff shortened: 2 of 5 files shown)


chrlesur:
Pre-commit adversarial review (`codex review`, default high-reasoning config; reviewer model: Codex, pilot model: Claude): no actionable regression — the cleanup is gated on the strict live evidence, every ambiguous case is pinned as keep by the 7-case matrix, and the deleted-device path no longer dereferences nil.

chrlesur:
Verdict: REQUEST-CHANGES (comment) — post-merge deep state-safety re-review finding, fix-forward in progress.

**[Blocking] The nil-drop path is not restricted to genuine deletions.**

The OpenIaaS API answers 403 for unknown AND forbidden ids alike, and the client maps it to a nil answer (`requireNotFoundOrOK(resp, 403)` — repo-wide convention, 51 call sites). With this PR's read-side cleanup, an access-denied response (expired token, IAM change) on `VirtualDisk().Read` or `NetworkAdapter().Read` would silently shrink `os_disk`/`os_network_adapter` in the state. Before this PR the same case crashed (nil dereference), so no behavior relied on it — but silent state shrink is worse than a crash.

**Fix (incoming PR)**: a nil per-id answer is only treated as a deletion after a second independent evidence — the device must also be absent from the VM-scoped `List` (which proves the token can read this VM's devices). If the List fails or still contains the id, the refresh fails closed with an explicit error.

Credit: the deep re-review also confirmed points 1, 2, 4 and 5 (strict predicate, clean plan convergence, warning quality, fail-closed update sequencing) — only the nil-path evidence was insufficient.

chrlesur:

> The patch introduces a state-safety regression by treating ambiguous nil API reads as confirmed deletions for nested OS devices. Because the client maps 403 to nil, reduced permissions can silently shrink Terraform state.$'\n'> $'\n'> Review comment:$'\n'> $'\n'> - [P1] Fail closed before dropping unreadable OS devices — /Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:637-645$'\n'>   The `Read` method used here returns `(nil, nil)` not only for a real 404 but also for 403 because it calls `requireNotFoundOrOK(resp, 403)`, so this drop path is not proof that the disk was deleted. If credentials lose access to a disk — and the same applies to the `os_network_adapter` nil check below — a refresh silently removes the nested device from Terraform state instead of failing closed, which can stop managing or recreate devices on a later plan. Please distinguish 404 from 403, or confirm deletion through a strict independent channel, before dropping state.$'\n'> The patch introduces a state-safety regression by treating ambiguous nil API reads as confirmed deletions for nested OS devices. Because the client maps 403 to nil, reduced permissions can silently shrink Terraform state.$'\n'> $'\n'> Review comment:$'\n'> $'\n'> - [P1] Fail closed before dropping unreadable OS devices — /Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:637-645$'\n'>   The `Read` method used here returns `(nil, nil)` not only for a real 404 but also for 403 because it calls `requireNotFoundOrOK(resp, 403)`, so this drop path is not proof that the disk was deleted. If credentials lose access to a disk — and the same applies to the `os_network_adapter` nil check below — a refresh silently removes the nested device from Terraform state instead of failing closed, which can stop managing or recreate devices on a later plan. Please distinguish 404 from 403, or confirm deletion through a strict independent channel, before dropping state.

chrlesur:
The blocking finding of the deep re-review is resolved by #275, merged into `rc/v1.8.0` on explicit human GO after the full strict review cycle (committed-diff independent review, verdict quoted verbatim).

PR #275, opened 2026-06-12
OpenIaaS: strict listing evidence before any state drop (fail closed on access-denied)

chrlesur:
Refs #273

Fix-forward of the blocking finding raised by the post-merge deep state-safety review of #274 (see the review trail there).

## The flaw

The OpenIaaS API answers **403 for unknown AND forbidden ids alike**, and the client maps both to a nil answer (`requireNotFoundOrOK(resp, 403)` — repo-wide convention, 51 call sites). The #274 nil-drop path could therefore turn an access-denied response (expired token, IAM change) into a silent state shrink. The first fix attempt — confirming through the regular VM-scoped `List` — had the **same circular flaw**: a forbidden listing also comes back nil and masquerades as an empty list (caught by the second adversarial pass).

## The fix

- New client variants `VirtualDisk().ListStrict` / `NetworkAdapter().ListStrict`: require a 2xx, surface an access-denied listing as an **error**.
- A nil per-id read is only treated as a deletion when the device is also absent from the **strict** VM-scoped listing (second independent evidence). If the listing fails or still contains the id, the refresh **fails closed** with an explicit error.
- The platform-managed config-drive cleanup is unaffected (it operates on a successfully read disk).

## Validation

- `TestDeviceConfirmedGone` (fail-closed matrix) + the existing 7-case `classifyOSDiskOnRead` matrix — green; full unit suites green.
- Two adversarial review iterations before this commit (both REQUEST-CHANGES verdicts incorporated), final pass: no actionable regression.


chrlesur:
Final adversarial pass (`codex review`, default high-reasoning config; reviewer model: Codex, pilot model: Claude): no actionable regression. Verdict trail on this topic: REQUEST-CHANGES on #274 (nil-drop not restricted to genuine deletions) → REQUEST-CHANGES on the first fix (circular evidence through the regular List) → this implementation with strict listing evidence.

chrlesur:
Final independent review of the COMMITTED branch state (diff vs `rc/v1.8.0`, `codex review`, default high-reasoning config; reviewer model: Codex). Reviewer conclusion, quoted verbatim:

> No actionable regression was found in the diff against the specified merge base. Targeted provider/client tests pass; the full `go test ./...` still has unrelated pre-existing PAT test signature failures outside this patch.

Process note: the previous comment on this PR wrongly carried a pilot-authored verdict line wrapping a pre-commit review — it must be disregarded as a formal verdict. From now on, the verdict published on a PR is exclusively the reviewer's own conclusion on the committed diff, and the merge is held for an explicit human GO on state-critical changes.

This PR is review-complete and awaiting the human merge decision.

PR #276, opened 2026-06-12
OpenIaaS: confirm the VIF cleanup before any retry, seed the id from success only

chrlesur:
Refs #251

FF-1 of the re-review campaign — fixes the P1/P2 findings raised on the merged #265 create path (per-PR re-review + full-train pass + targeted deep review, see the verdict trail on #265).

## Findings addressed

1. **[P1] Unconfirmed cleanup no longer allows a retry** — a refused delete, a failed delete activity or an unavailable strict listing now aborts with an explicit error: retrying after an unconfirmed cleanup could duplicate the VIF on the VM.
2. **[P2] The resource id is only seeded from a successful activity** — previously `setIdFromActivityState` also ran on failed activities, so the state could point at the adapter the cleanup had just deleted. An empty id after a malformed success now aborts before the `attached`/Disconnect block.
3. **Ownership proof before deletion** — only adapters referenced by THIS create's failed activity AND confirmed present on THIS VM by `ListStrict` are deleted (intersection = existence + ownership evidence; protects against the race where another actor creates an adapter on the same VM).
4. Context cancellation surfaces `ctx.Err()` (create loop + update wrapper).

## Review cycle (strict)

- Targeted attack review on the uncommitted diff: first pass found a 5th-invariant gap (empty-id Disconnect) → fixed → second pass **VERDICT: VALIDATED** (with the ownership-proof analysis documented).
- Committed-diff independent review: to follow on this PR, verdict quoted verbatim.
- Merge held for explicit human GO (state-critical zone).

## Validation

6 new unit cases (`TestVIFCleanupTargets` matrix, cancellation propagation) + all suites green.


chrlesur:
Verdict: REQUEST-CHANGES (comment)

> The patch hardens several retry paths, but the create cleanup still allows a retry after a referenced half-created adapter was not confirmed/deleted. That leaves the duplicate-adapter risk unresolved in a concrete transient create-failure case.
> 
> Review comment:
> 
> - [P1] Abort retry when cleanup target is unconfirmed — /Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_network_adapter.go:327-327
>   When the failed create activity references a `network_adapter` but the immediate strict VM listing does not include that id, `vifCleanupTargets` returns no targets and `cleanupHalfCreatedVIFs` returns success, so the create loop retries without deleting or aborting. In the transient create-failure scenario this can still duplicate a half-created VIF if the listing is stale or incomplete; the unconfirmed referenced adapter should make the cleanup fail closed rather than be treated as a confirmed cleanup.
> The patch hardens several retry paths, but the create cleanup still allows a retry after a referenced half-created adapter was not confirmed/deleted. That leaves the duplicate-adapter risk unresolved in a concrete transient create-failure case.
> 
> Review comment:
> 
> - [P1] Abort retry when cleanup target is unconfirmed — /Users/clesur/PROJETS/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_network_adapter.go:327-327
>   When the failed create activity references a `network_adapter` but the immediate strict VM listing does not include that id, `vifCleanupTargets` returns no targets and `cleanupHalfCreatedVIFs` returns success, so the create loop retries without deleting or aborting. In the transient create-failure scenario this can still duplicate a half-created VIF if the listing is stale or incomplete; the unconfirmed referenced adapter should make the cleanup fail closed rather than be treated as a confirmed cleanup.

Fix incoming on this branch; a referenced-but-unlisted adapter will fail closed instead of allowing the retry.

chrlesur:
Independent review of the COMMITTED diff vs `rc/v1.8.0` after the hardenings (`codex review`, gpt-5.5 high default config). Reviewer conclusion, quoted verbatim:

> I did not identify any actionable regression introduced by this patch. The targeted provider and client unit tests pass locally with a sandboxed Go build cache.
> I did not identify any actionable regression introduced by this patch. The targeted provider and client unit tests pass locally with a sandboxed Go build cache.

Review cycle on this PR: committed-diff REQUEST-CHANGES (referenced-but-unlisted retry hole) -> fail-closed fix -> two further attack iterations (silent activity, cardinality, top-of-loop ctx) -> attack VERDICT: VALIDATED -> this committed-diff pass. Merge held for explicit human GO.

PR #277, opened 2026-06-12
OpenIaaS: restore the power state on failed updates, gate the residual Optional+Computed properties

chrlesur:
Refs #267

FF-2 of the re-review campaign — fixes the deep-review findings on the Lot A zone (see the verdict trail on #268).

## Changes

1. **No more VMs left powered off**: every error path between a provider-initiated power-off and the normal power-on (properties PATCH failure, disk/adapter update failures, uncertain power-off outcome) attempts a restore and reports the outcome in the error. The restore helper reads the live power state first (already-running = success, never a double power-on) and targets the **pre-shutdown host** (live evidence).
2. **Residual Optional+Computed gating**: `num_cores_per_socket` and `boot_firmware` join `secure_boot` as raw-config-gated pointers — an unconfigured value is never pushed even when the state-merged value diverges from live (stale state must not overwrite a live setting). Explicit `num_cores_per_socket = 0` is rejected at plan time.

## Review cycle (strict)

Four attack-review iterations before commit — each one produced a real hardening (restore on uncertain power-off; idempotent state-checking restore; zero-cores plan-time rejection; pre-shutdown host evidence) — final attack verdict **VALIDATED**. Committed-diff independent review to follow on this PR, verdict quoted verbatim.

## Validation

Updated `TestBuildOpenIaasVMPropertiesPatch` matrix (stale-state non-push pinned, configured-divergent firmware pushed); all unit suites green.


chrlesur:

> No actionable regressions were found in the diff. The targeted and workflow-equivalent Go tests for internal/client, internal/provider, and internal/provider/helpers pass locally.
> No actionable regressions were found in the diff. The targeted and workflow-equivalent Go tests for internal/client, internal/provider, and internal/provider/helpers pass locally.

PR #278, opened 2026-06-12
Client: not-found tolerance independent from the transient retry budget

chrlesur:
Refs #245

FF-3 of the re-review campaign — fixes the P2 finding of the #249 per-PR re-review.

## Change

The one-time tolerance for an activity not yet visible (eventual consistency right after the operation starts) was consumed by ANY prior polling attempt: a transient 429/5xx/transport blip before the first successful read turned the first legitimate not-found into a permanent failure. The tolerance is now an independent flag (`notFoundTolerated`), and the disappearance of an activity that was already seen stays permanently fatal (`activitySeen`).

## Review cycle

Attack review on the uncommitted diff (interleavings, unbounded-polling check, count-removal side effects, GC'd-activity semantics): **VERDICT: VALIDATED** first pass. Committed-diff review to follow, verdict verbatim.

## Validation

2 new pinned cases (transient blip before first not-found; seen-then-vanished permanent) + the 8 existing invariant groups — all green.


chrlesur:

> The activity polling change cleanly separates the one-time not-found tolerance from transient read retry accounting while preserving the permanent failure path for activities that were already observed and then disappear. The targeted client tests pass with a sandbox-local Go cache.
> The activity polling change cleanly separates the one-time not-found tolerance from transient read retry accounting while preserving the permanent failure path for activities that were already observed and then disappear. The targeted client tests pass with a sandbox-local Go cache.

PR #279, opened 2026-06-12
VMware: gate every boot_options attribute on explicit configuration

chrlesur:
Refs #271

FF-4 of the re-review campaign — fixes the Zone C deep-review finding (see the verdict trail on #272).

## Changes

- Every `BootOptions` attribute is raw-config-gated: `boot_delay`/`boot_retry_delay` become `*int,omitempty` (an explicit `0` stays expressible), `firmware` becomes omitempty **and validated as `bios`/`efi` at plan time** (an explicit empty value can no longer be silently dropped by the serializer), the three booleans were already gated.
- A present block with **no configured attribute sends nothing**: an empty `bootOptions` object could reset live values platform-side.

## Review cycle

Attack review pass 1: two P1 findings (silent empty-firmware drop; empty-block payload) → fixed → pass 2 **VERDICT: VALIDATED**. Committed-diff review to follow, verdict verbatim.

## Validation

Updated `TestBuildVMwareBootOptionsFromRaw` matrix (unconfigured attributes omitted, explicit zero carried, unconfigured block = nil payload); all unit suites green.


chrlesur:

> No actionable regression was found in the diff. The provider checks passed with a sandbox-local Go cache; the broader client test failure observed is in unrelated PAT tests outside this patch.
> No actionable regression was found in the diff. The provider checks passed with a sandbox-local Go cache; the broader client test failure observed is in unrelated PAT tests outside this patch.

PR #280, opened 2026-06-12
OpenIaaS: accept a deletion only under strict, complete listing evidence

chrlesur:
Refs #271

FF-5 — the last finding batch of the re-review campaign (Zone C). **State-shape critical: merge held for explicit human GO.**

## The doctrine, now complete on OpenIaaS

A nil per-id read (the API answers 403 for unknown AND forbidden ids alike) is only a deletion when a **strict listing** confirms it:
- `ListStrict` variants require a **complete 200** answer (a 206 partial listing cannot prove an absence; an access-denied listing is an error);
- the tri-state verdict (`classifyMissingDevice`, unit-tested): absent everywhere = deletion confirmed; still in the scoped listing = fail closed (access restriction); **present tenant-wide but detached/moved = drift, never a deletion** (actionable error);
- applied to the four OpenIaaS resources: virtual machine (top-level Read included), virtual disk, standalone network adapter (whose Read still had the raw nil-drop), replication policy.

## Review cycle

Three attack iterations, each with real findings (VM-scoped-only evidence too narrow; standalone adapter missed; 206 acceptance; top-level VM) → **VERDICT: VALIDATED**. Committed-diff review to follow, verdict verbatim. Residual nil-drops outside OpenIaaS (VMware resources, object storage, PAT) are catalogued for a dedicated sweep — next item of the remediation series.

## Validation

`TestClassifyMissingDevice` + full unit suites green.


chrlesur:

> The diff tightens deletion evidence for OpenIaaS resources and the touched unit tests compile and pass with `go test ./internal/client ./internal/provider`. I did not find an introduced, actionable regression in the changed lines.
> The diff tightens deletion evidence for OpenIaaS resources and the touched unit tests compile and pass with `go test ./internal/client ./internal/provider`. I did not find an introduced, actionable regression in the changed lines.

Merge held for explicit human GO (state-shape critical zone).

Issue #241, opened 2026-05-20
Error unmarshalling data.cloudtemple_compute_virtual_machine


NBardelot:
## Describe the bug
When I retrieve a *cloudtemple_compute_virtual_machine* as a data, the provider throws the following error:
```
╷
│ Error: virtual_machines.0.extra_config: '': source data must be an array or slice, got map
│
│   with data.cloudtemple_compute_virtual_machines.all,
│   on data.tf line 1, in data "cloudtemple_compute_virtual_machines" "all":
│    1: data "cloudtemple_compute_virtual_machines" "all" {}
│
╵
```
When debugging the provider (`export TF_LOG_PROVIDER=DEBUG`) I see that the Received HTTP Response does not contain any  `extra_config`.

## Current Behavior
The error is systematic. It happens even for VMs created very recently.

## Expected behavior
The data should be unmarshalled correctly.

## Steps to reproduce
A simple `test.tf`:
```
terraform {
  required_providers {
    cloudtemple = {
      source  = "Cloud-Temple/cloudtemple"
      version = "1.7.1"
    }
  }
}

provider "cloudtemple" {}

data cloudtemple_compute_virtual_machines "all" {}
data cloudtemple_compute_virtual_machine "all" {
  for_each = data.cloudtemple_compute_virtual_machines.all.virtual_machines
  id = each.key
}
```

The error also happens when trying to use version 1.6.1 of the provider.

PR #287, opened 2026-06-14
fix(#241): expose VMware datasource extra_config as a map

chrlesur:
## Problem

The datasources `cloudtemple_compute_virtual_machine` and `cloudtemple_compute_virtual_machines` fail **systematically** at read time (reported since provider 1.6.1), making both unusable:

```
Error: virtual_machines.0.extra_config: '': source data must be an array or slice, got map
```

## Root cause

The shared `FlattenVirtualMachine` helper (used by the resource **and** both datasources) emits `extra_config` as a `map[string]string`. The **resource** declares `extra_config` as `TypeMap` (canonical shape, with a `StateUpgrader` that migrates the old list format to a map) and works. Both **datasource** schemas declared `extra_config` as a `TypeList` of `{key, value}` objects, so `d.Set` receives a map for a list field and fails — the singular per-key `Set` and the plural single `Set` both abort the read.

## Fix

- Align `extra_config` to `TypeMap(String)` in **both** datasource schemas, matching the resource and the actual API shape.
- The shared `FlattenVirtualMachine` helper and the resource (`extra_config`, `DiffSuppressFunc`, `StateUpgrader`, `ConvertExtraConfigValue`) are **left untouched** — the resource already relies on map semantics.

## Tests (non-complacent) + CI

- Extends the CI schema-vs-flatten walker `TestDatasourceFlattenOutputsFitTheirSchemas` to cover **both** VMware VM datasources with a fully-populated `fixtureVMwareVirtualMachine()` (every nested block exercised; the plural check mirrors the real read by adding `id`). This is the exact mechanism that already guards the OpenIaaS datasources (#243 class) — it did not cover the VMware VM datasources, which is why #241 slipped through.
- Verified the walker **fails on the bug**: reverting the schema to the old `TypeList` reproduces `extra_config: source data must be an array or slice, got map`.
- `gofmt`, `go vet ./...`, `go build ./...` and `go test ./internal/client ./internal/provider ./internal/provider/helpers` all green.

## Docs

Regenerated the two datasource docs (`extra_config`: *List of Object* → *Map of String*).

## State safety

Read-only datasources: no `StateUpgrader` and no managed-resource state migration is involved; this cannot shrink or corrupt any managed resource state. The datasource has been unreadable since at least 1.6.1, so no working usage relies on the old list shape.

## Upgrade note

A configuration that referenced the previous (never-readable) list shape (e.g. `extra_config[0].key`) must switch to map access (e.g. `extra_config["disk.enableUUID"]`). Such a reference could not have worked before this fix; a `try()` wrapper around it would now resolve differently.

Refs #241


chrlesur:
## Review (Codex gpt-5.5, high, fast)

Two passes were run against base `rc/v1.8.0`.

**1. Canonical committed-diff review** (`codex review --base rc/v1.8.0`): ran the targeted and package-level provider tests (green), no whitespace/conflict issues. Conclusion: the schema change aligns the VMware datasource declarations with the actual flatten output shape, and the added regression coverage exercises both singular and plural datasources.

**2. Adversarial deep review** (attack points A1–A7: test non-complacency, fixture faithfulness vs both real Read paths, zero-value hiding, state safety, empty-map edge, scope/regression, other plan/apply/refresh risks):

- **A1 Non-complacent test**: the walker would fail if `extra_config` regressed to `TypeList`; undeclared keys `t.Errorf` (not skipped), and the fixture uses a non-empty `ExtraConfig` so it is not hidden by an empty map.
- **A2 Fixture faithfulness**: the plural fixture correctly mirrors the real Read adding `id`; the singular does not (matches reality). The walker proves every key `FlattenVirtualMachine` emits is declared by the singular schema.
- **A3 Zero-value hiding**: the three conditional nested blocks (`replication_config`, `storage`, `boot_options`) are all set non-zero in the fixture; no other conditionally-emitted key; `extra_config.value` is always a string.
- **A4 State safety**: no managed-resource migration, no `StateUpgrader` needed; cannot shrink/corrupt managed-resource state. Old datasource reads failed loudly; the upgrade is an acceptable datasource state-shape rewrite.
- **A5 Empty extra_config**: `TypeMap` accepts an empty (non-nil) map; the old `TypeList` rejected any map regardless of length. No nil-vs-empty diff issue.
- **A6 Scope**: clean — two datasource schemas, the walker test, docs and changelog only; no dangling doc anchors.
- **A7**: one non-blocking caveat — the list→map expression-shape change is breaking for anyone referencing the old (never-readable) shape; covered by the CHANGELOG upgrade note.

> REVIEW VERDICT: GO — the fix is narrowly scoped to the datasource schema mismatch, and the added walker coverage would catch a regression back to `TypeList`.

No blocking finding. Awaiting CI green and the human merge decision (state-shape change).

Issue #288, opened 2026-06-14
Extend the datasource schema-vs-flatten walker to every datasource


chrlesur:
## Context

The CI unit walker `TestDatasourceFlattenOutputsFitTheirSchemas` (`internal/provider/datasource_flatten_schema_test.go`) sets each datasource's flatten output into its real schema through the SDK writer and fails on any shape/type mismatch. This guards against the systematic read-breaking class where a flatten helper emits a key (or shape) the schema does not accept — the bug class behind **#243** (`Invalid address to set`) and **#241** (`source data must be an array or slice, got map`).

Today the walker only covers the **26 OpenIaaS datasources** plus the **2 VMware VM datasources** (added with #241). Every other datasource (VMware compute, backup, IAM, object storage, marketplace) is **unguarded** and may hide a latent #241/#243-class systematic failure.

An inventory confirms that **every datasource uses a `helpers.Flatten*` helper**, so the walker mechanism applies uniformly.

## Goal

Bring the walker to **100% datasource coverage**:

- one fully-populated fixture per distinct client struct (every field non-zero, so type/shape mismatches cannot hide behind zero values — non-complacent);
- one check per datasource (singular with empty `rootKey`; plural with its computed list attribute, mirroring any post-flatten key the real Read injects, e.g. `id`);
- any latent same-class read-breaking mismatch revealed is fixed in scope (align the schema to the real flatten/API shape — never weaken the flatten helper or the test), with a CHANGELOG entry and regenerated docs;
- broader design drift (if any) is split out rather than handled here.

This is the test-coverage hardening mandated for the read path; it is platform-independent and runs in the existing CI `unit` job.

Refs #264, #243, #241

chrlesur:
Delivered the walker extension via #289 (merged into `rc/v1.8.0`): the schema-vs-flatten CI validation now covers every datasource (reflection-based non-zero filler + coverage registry + a meta-test that fails if a datasource is neither covered nor a declared known gap).

`cloudtemple_iam_features` is recorded as an explicit **known gap**: pushing the walker to full recursive depth showed that `FlattenFeature` can emit `subfeatures` beyond the schema's declared nesting (a tree deeper than the declared depth breaks the read — verified at depth 4) and that the swagger (`/v2/features` flat) contradicts the client model and the acceptance test. The coherent fix (schema/model/flatten/test, informed by the real API contract) is handled as a separate effort and will reference this issue.

[...]

PR #289, opened 2026-06-14
test(#288): extend the datasource schema-vs-flatten walker to (almost) every datasource

chrlesur:
## Context

The CI unit walker `TestDatasourceFlattenOutputsFitTheirSchemas` sets each datasource's flatten output into its real schema through the SDK writer and fails on any shape/type mismatch — the systematic read-breaking class behind **#243** (`Invalid address to set`) and **#241** (`source data must be an array or slice, got map`). It only covered the 26 OpenIaaS datasources + the 2 VMware VM datasources; every other datasource was **unguarded**.

## Change (test-only)

- **Reflection non-zero filler** (`filled[T]`): fully populates a client object (strings, numbers, bools, slices, maps, pointers, arrays, nested/anonymous structs, `time.Time`), so a flatten/schema mismatch cannot hide behind a zero value and coverage grows automatically when a client struct gains a field. Recursion is depth- and self-reference-bounded.
- **Coverage registry**: maps every datasource to its flatten payload, mirroring each real Read exactly (`rootKey`; injected `id` where the Read injects it; empty `vmID` for the OpenIaaS virtual disk; the multi-section `backup_metrics`).
- **Meta-test** `TestEveryDatasourceIsCoveredByTheFlattenWalker`: every registered datasource must be either covered or a **declared known gap** (`datasourceKnownGaps`), never both, and neither map may reference an unregistered datasource. Proven to fail in both directions.

## Scope: walker-only; `cloudtemple_iam_features` is a declared known gap

Pushing the walker to full recursive depth surfaced that `iam_features` is genuinely inconsistent: `FlattenFeature` can emit `subfeatures` beyond the schema's declared nesting (a tree deeper than the declared depth crashes the read — verified at depth 4), **and** the swagger (`/v2/features` flat) contradicts the client model and the acceptance test. The correct shape is undetermined without a real API response. Rather than ship a partial, possibly-moot fix, `iam_features` is an **explicit declared known gap** here and its coherent fix (schema/model/flatten/test, informed by the real contract) is handled separately. `FlattenFeature` is unchanged from `rc/v1.8.0`.

Net diff: the walker test file + one CHANGELOG line. No production/runtime change, no state-shape impact.

## Reviews
- Codex plan review: GO (reflection filler + override/known-gap approach).
- Codex diff review: initial **NO-GO** (the deeper walker surfaced the `iam_features` inconsistency) → scope reduced to walker-only with the declared gap → **GO** (walker-only net diff, explicit known gap, no runtime/state-shape change).

Refs #288, #241, #243, #264


chrlesur:
## Review (Codex gpt-5.5, high, fast) — adversarial, two rounds

**Round 1 — NO-GO.** The reflection filler's self-reference cap left `cloudtemple_iam_features` under-filled: only one nested level was exercised while the schema declares two (`features[].subfeatures[].subfeatures[]`), so the deepest declared recursive level was not exercised — complacent. The review also pointed at `FlattenFeature` always emitting an (even empty) `subfeatures` key.

This turned out to be a **real #243-class latent bug**: a leaf feature at the deepest declared level still emitted an empty `subfeatures` key, which the deepest schema element does not declare, so the `cloudtemple_iam_features` read would break with `Invalid address to set` for sufficiently nested feature trees.

**Fix applied (in-scope, same class):**
- `FlattenFeature` now emits `subfeatures` only when sub-features exist (read-only datasource; not shared with any resource).
- The walker covers `iam_features` with a hand-built tree reaching the deepest declared level — **proven non-complacent**: with the empty-`subfeatures` behavior restored, the check fails with exactly `Invalid address to set: features.0.subfeatures.0.subfeatures.0.subfeatures`; with the fix it passes.
- Hardened the reflection filler (`reflect.Array`) and `flatID` (panics on misuse instead of silently skipping the id).

**Round 2 — re-review verdict (verbatim):**

> REVIEW VERDICT: GO

Codex confirmed: the NO-GO is resolved (deepest `iam_features` level now exercised non-complacently); the `FlattenFeature` fix is correct and state-safe (read-only datasource; a leaf's `subfeatures` goes from `[]` to absent, harmless for a Computed list; strictly improves robustness up to the schema's declared depth); the filler has no material remaining gap; fidelity and the meta-test are intact; the diff is test + the single read-only flatten fix + CHANGELOG. Residual (assumed): feature trees deeper than the schema's declared levels remain bounded by the schema — a separate schema-extension concern, not reintroduced here.

No blocking finding. Awaiting CI green and the human merge decision.

chrlesur:
## Final review (Codex gpt-5.5, high, fast) after scope reduction

A parallel investigation showed the earlier `FlattenFeature` change was only a **partial** fix (a feature tree deeper than the schema's declared nesting still crashes — verified at depth 4) and that the swagger (`/v2/features` flat) contradicts the client model + acceptance test. The features fix was therefore **reverted** and `cloudtemple_iam_features` made an **explicit declared known gap**; the coherent fix is handled separately. `FlattenFeature` now has zero diff vs `rc/v1.8.0`.

Codex confirmed on the reduced diff: net diff is walker test + one CHANGELOG line (no production/runtime change, no state risk); the known gap is enforced by the meta-test (explicit debt, not a silent pass; `iam_features` is not executed by the walker); the remaining ~89 datasources stay non-complacently covered; CHANGELOG is accurate.

> REVIEW VERDICT: GO — walker-only net diff, explicit known gap, no runtime/state-shape change.

Awaiting CI green and the human merge decision.

PR #290, opened 2026-06-14
fix(#288): bound iam_features flatten to the schema's declared depth

chrlesur:
Refs #288, #243

## Problem

The `cloudtemple_iam_features` datasource carried a latent #243-class bug. `FlattenFeature` always emitted a `subfeatures` key, but the schema declares a finite `subfeatures` nesting whose deepest element declares only `{id, name}`. A feature tree deeper than that depth — a node at the deepest declared level that itself has children, i.e. a **non-empty** list under an undeclared key — made `d.Set` fail with `Invalid address to set`, which makes the **whole datasource unusable** (read-breaking, TF prod breaker class).

The real catalog depth is **2** (verified live against `GET /iam/v2/features`, IAM API v2.47.0: only `rtms` has sub-features, both leaves), so current outputs are unaffected; this closes the crash for any deeper tree the API might return. The IAM swagger is incomplete — it does not document `subFeatures` although the API returns it.

## Fix

- `FlattenFeature` is now **depth-bounded** to the schema's declared nesting: it emits `subfeatures` only while the schema declares it (keeping the empty list at declared levels, so the **state shape is unchanged** for existing data), and **truncates** beyond the deepest declared level instead of crashing. Nil sub-features are skipped; a nil feature returns nil.
- `dataSourceFeaturesRead` emits a `Warning` (with the feature name/id) when a sub-tree is truncated, so deeper data is never dropped silently.
- The schema is **unchanged** (3 levels), so the generated docs are unchanged.

## Tests (non-complacent)

- The schema-vs-flatten walker now **covers** `cloudtemple_iam_features` (graduated from a known gap) with a hand-built depth-4 tree.
- `TestIAMFeaturesFlattenIsDepthBounded` and the helper tests prove the read no longer breaks; they **fail** on the unbounded helper with the exact `Invalid address to set: features.0.subfeatures.0.subfeatures.0.subfeatures`.
- `TestIAMFeaturesSchemaAndFlattenDepthAgree` introspects the schema and asserts the flatten bound matches the declared nesting, guarding against future drift.
- The acceptance test `TestIAM_Features` is enriched to lock the observed depth (sub-features of `rtms` are leaves) without weakening it.

CHANGELOG updated (BUG FIXES); the MISCELLANEOUS walker line no longer excludes `cloudtemple_iam_features`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ BUG FIXES :
   * Fixed error paths that could leave a virtual machine powered off after a provider-initiated shutdown on resource `cloudtemple_compute_iaas_opensource_virtual_machine`: the provider now restores the power state on failure (best effort, targeting the pre-shutdown host) and reports the outcome in the error.
   * Fixed Terraform states polluted by pre-1.8.0 creations that captured the cloud-init config drive (`XO CloudConfigDrive`) as a managed `os_disk`: the refresh now drops the drive from the state — with an explicit warning and only under strict live evidence (exact XO name and read-only attachment to this virtual machine) — which stops the permanent removal drift for existing clients. Ambiguous disks stay in the state. No platform-side change is ever made.
   * Fixed a crash (nil dereference) during refresh when an `os_disk` or `os_network_adapter` referenced in the state no longer exists platform-side: the deleted device is now dropped from the state so the next plan reflects reality. The deletion is confirmed by a second independent read (the VM-scoped device listing) before any state entry is dropped, so an access-denied answer can never silently shrink the state.
+  * Fixed datasource `cloudtemple_iam_features` being able to fail at read time with an `Invalid address to set` error, which would make the whole datasource unusable, if the feature tree were ever nested deeper than the schema declares (a sub-feature below the deepest declared level emitting a `subfeatures` key the schema does not declare). The flatten is now bounded to the schema's declared nesting depth: any deeper sub-features are truncated — the datasource stays readable — and a warning is emitted, instead of breaking the read. The real feature catalog is only two levels deep, so current outputs are unchanged.
 
 IMPROVEMENTS :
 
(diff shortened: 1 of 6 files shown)


chrlesur:
## Pilot review (independent verification + Codex)

I verified this PR independently (not on the agent's word).

**Local verification (review worktree):**
- Full suite green (`./internal/client ./internal/provider ./internal/provider/helpers`); the 5 helper tests + the 2 provider-level feature tests pass.
- **Non-complacency proven**: reinjecting an unbounded `FlattenFeature` (removing the depth guard) makes all four discriminating tests FAIL — `TestFlattenFeatureIsDepthBounded` (level-2 carries a `subfeatures` key), `TestDatasourceFlattenOutputsFitTheirSchemas/cloudtemple_iam_features` and `TestIAMFeaturesFlattenIsDepthBounded` (`Invalid address to set`), and `TestIAMFeaturesSchemaAndFlattenDepthAgree` (emits 4 levels vs 2 declared). The tests genuinely guard the bug.

**Assessment:** the fix preserves the exact output shape for the real API depth (2) — no drift for existing consumers — and converts the latent depth≥4 crash into bounded truncation + an actionable `diag.Warning`. Schema/docs unchanged. The schema-introspection test ties the flatten bound to the declared depth (anti-drift). Scope is contained (helper, Read, walker entry, 2 tests, CHANGELOG). Coordination with #289 is consistent (`iam_features` graduated from known-gap to coverage; gaps map empty; meta-test intact).

**Codex (gpt-5.5, high, fast), committed-diff adversarial review — verbatim verdict:**

> REVIEW VERDICT: GO — the real API state shape is preserved exactly while the latent crash path is converted into bounded truncation with a matching warning.

**Governance:** all fields complete. One classification correction: **TF prod breaker RED → NO** — the crash is unreachable with the live API (depth 2, never failed) and the fix makes even a hypothetical deeper tree non-breaking; `TF prod breaker` reflects real production impact, which is NO here.

No blocking finding. Awaiting the human merge decision.

PR #292, opened 2026-06-14
ci(#264): gate the unit job on staticcheck and the race detector

chrlesur:
## What

Increment 2 of the platform-independent CI safety net (#264). The minimal net (#266) already runs gofmt + `go vet ./...` + unit tests with coverage. This PR adds two static guards to the `unit` job, both running **without any live platform**:

- **staticcheck** — pinned: `go run honnef.co/go/tools/cmd/staticcheck@2025.1.1 ./...` (supports Go 1.24; analyses the whole module, incl. the acceptance-test packages that are compiled but not run).
- **race detector** — the pure unit tests now run under `-race`.

## staticcheck findings cleared (6, none state-shape)

| Check | Location | Fix |
|---|---|---|
| ST1005 | `internal/client/api.go` | lowercase the `access denied` error string; the acceptance test pinning it is updated to match |
| SA1006 | `…/data_source_compute_guest_operating_system_test.go` | drop the pointless `fmt.Sprintf` around a constant |
| U1000 | `internal/provider/provider.go` | remove unused `getUserID` |
| U1000 | `internal/provider/tags.go` | remove unused `readTags` |
| U1000 | `…/resource_compute_virtual_machine_unit_test.go` | remove unused `bootOptionsRawType` var |
| SA4006 | `internal/client/tests/compute_virtual_machine_test.go` | discard the dead `activityId` store |

## State safety

None of the 6 fixes change Terraform-state semantics. The `api.go` change is **case-only** in the human-facing error text; no retry/error-classification logic compares that string (classification uses typed errors). The three removed symbols are unreferenced across the module.

## Honest scoping

- `-race` is framed as a **future-proofing guard** (token cache, httptest servers, timers), not state-safety coverage. staticcheck is the main win here.
- **Known follow-up (not in this PR):** the SA4006 site reveals a latent acceptance-test defect — `AssignVirtualMachine`'s activity is never `WaitForCompletion`'d (analogous call sites do). Discard is the faithful minimal CI-gate fix; the wait is tracked separately.

## Out of scope (later #264 increments)

Marketplace datasource tests, broader pure-logic unit layer, schema-change detection, plan-convergence checks.

## Local verification

`gofmt -l .` empty · `go vet ./...` clean · `staticcheck ./...` **0 findings** · `go build ./...` ok · `go test -race -count=1 -cover ./internal/client ./internal/provider ./internal/provider/helpers` green (cov 10.8 / 8.5 / 12.9 %).

Refs #264

PR #294, opened 2026-06-14
test(#264): await the SLA-policy assignment activity in the VM acceptance test

chrlesur:
## What

`TestCompute_UpdateAndPower` in `internal/client/tests/compute_virtual_machine_test.go`
called `client.Backup().SLAPolicy().AssignVirtualMachine(...)` but discarded the
returned activity id and never awaited the activity. This change captures the
activity id and waits for its completion, matching the surrounding pattern:

```go
activityId, err = client.Backup().SLAPolicy().AssignVirtualMachine(ctx, &clientpkg.BackupAssignVirtualMachineRequest{
    VirtualMachineIds: []string{instanceId},
    SLAPolicies:       []string{os.Getenv(PolicyId)},
})
require.NoError(t, err)
_, err = client.Activity().WaitForCompletion(ctx, activityId, nil)
require.NoError(t, err)
```

## Why

Every analogous call site awaits the activity it triggers, and so does the
production code:
- `internal/client/tests/backup_sla_policy_test.go:93`
- `internal/client/tests/compute_network_adapter_test.go:91`
- `internal/provider/resource_compute_virtual_machine.go:1442`

Without the wait, the test moves on before the SLA-policy assignment is
confirmed applied, which can race the subsequent `VirtualMachine().Read`.

This supersedes the SA4006 dead-store discard introduced in 22f2cc5 (#292):
the activity id is now used rather than thrown away, so `staticcheck ./...`
stays clean.

## Scope / risk

- Touches a single acceptance test (`TF_ACC=1`, requires a live Cloud Temple
  platform); it does not run in CI. No state-shape change, no production code
  change.
- `go vet ./...` and `staticcheck ./...` (pinned 2025.1.1) are clean locally.

Refs #264

### DIFF
--- a/internal/client/tests/compute_virtual_machine_test.go
+++ b/internal/client/tests/compute_virtual_machine_test.go
@@ -108,10 +108,12 @@ func TestCompute_UpdateAndPower(t *testing.T) {
 	_, err = client.Backup().Job().WaitForCompletion(ctx, jobs[0].ID, nil)
 	require.NoError(t, err)
 
-	_, err = client.Backup().SLAPolicy().AssignVirtualMachine(ctx, &clientpkg.BackupAssignVirtualMachineRequest{
+	activityId, err = client.Backup().SLAPolicy().AssignVirtualMachine(ctx, &clientpkg.BackupAssignVirtualMachineRequest{
 		VirtualMachineIds: []string{instanceId},
 		SLAPolicies:       []string{os.Getenv(PolicyId)},
 	})
+	require.NoError(t, err)
+	_, err = client.Activity().WaitForCompletion(ctx, activityId, nil)
 	require.NoError(t, err)
 
 	vm, err := client.Compute().VirtualMachine().Read(ctx, instanceId)


chrlesur (commented):
## Independent review — Codex (read-only, role distinct from the pilot)

Reviewer role held by Codex (codex-cli 0.139.0), separate from the pilot model that authored the change. Published verbatim below.

---

REVIEW VERDICT: GO

No blocking findings.

Verified against `origin/rc/v1.8.0...HEAD`: the PR changes only `internal/client/tests/compute_virtual_machine_test.go` (+3/-1). `activityId` is already declared in `TestCompute_UpdateAndPower` at line 74, so the new `activityId, err = ...` at line 111 is a valid reassignment and does not shadow.

The wait pattern is consistent with the repository:
- `compute_virtual_machine_test.go:111-117` now captures the SLA assignment activity and waits on `client.Activity().WaitForCompletion(ctx, activityId, nil)`.
- `compute_network_adapter_test.go:91-98` and `backup_sla_policy_test.go:93-100` already use the same client-test pattern.
- Production code does the same in `resource_compute_virtual_machine.go:1442-1452` for update, and also in create at `1074-1084`.

Regression/order risk: I do not see a new invariant violation. The wait is inserted immediately after the async SLA assignment and before the VM is read/updated/powered, which makes the test ordering stricter and closer to production behavior. The only expected cost is that the live test may run longer or fail earlier if the assignment activity fails, which is the point of the fix.

State/Terraform impact: none. This is test-only, with no schema, provider, resource, state upgrader, or production client behavior change. One nuance: I do not see `TF_ACC` gating in `internal/client/tests`; `TestMain` directly requires `SHIVA_TEST_*` env vars. These tests are not executed by the active CI test job, but the workflow does compile/analyze test packages through `go vet ./...` and staticcheck.

Staticcheck/vet: I could not complete local `go vet`/`go test` because the read-only sandbox prevents Go from creating its temporary build directory. The local `staticcheck` binary returned `matched no packages`, so I did not rely on that result. By inspection, the specific SA4006 site introduced by `22f2cc5` is fixed: the activity ID returned by `AssignVirtualMachine` is no longer discarded or overwritten before use; it is consumed immediately by `WaitForCompletion`.

Non-blocking pre-existing note: the test still waits for `jobs[0].ID` after running the selected `job.ID`, while production waits on the selected `job.ID`. That fragility predates this PR and is not worsened by the added SLA assignment wait.

---

**Pilot note (covering the one item Codex could not run inside its read-only sandbox):** `go vet ./...` and `staticcheck ./...` (pinned 2025.1.1, identical to the CI gate added in #292) were both run locally on this branch and exited 0 — SA4006 no longer fires. All three GitHub Actions checks on this PR currently report pass (Build, Unit tests & vet, generate).

Issue #293, opened 2026-06-14
Non-complacent unit-test coverage for the pure / state-facing layer


chrlesur:
## Summary

Establish **non-complacent unit-test coverage** for the provider's pure, state-facing logic, everywhere it is currently absent. This is the workstream-level tracker for lot **S7 Test coverage**. It continues the unit-test layer scoped in #264, but as a dedicated, exhaustive program rather than an opportunistic add-on.

**Non-complacency definition (binding):** a test must *fail on the real bug*. For every pure function under test, the suite must red out if the function is mutated to the historical failure mode (wrong key emitted, nil panic, zero-vs-absent confusion, dropped list element, secret overwritten, wrong error class). Happy-path-only tests do not count.

## Coverage baseline (measured 2026-06-14, pure packages, no live platform)

| Package | Total coverage | Functions at 0% |
|---|---|---|
| `internal/client` | 10.8% | 258 / 295 |
| `internal/provider` | 8.5% | 159 / 290 |
| `internal/provider/helpers` | 12.9% | 72 / 79 |

Not all 0% functions are unit-testable: most `internal/client` functions are thin HTTP wrappers exercised only by the live acceptance suite (#8). This program targets the **pure logic** subset.

## Progress (`rc/v1.8.0` @ `20239c0`)

Coverage floor raised since the 2026-06-14 baseline:

| Package | Baseline | Current floor |
|---|---|---|
| `internal/client` | 10.8% | 21.0% |
| `internal/provider` | 8.5% | **16.1%** |
| `internal/provider/helpers` | 12.9% | **68.1%** |

## Enforcement: coverage ratchet (CI)

A per-package coverage floor is recorded and checked in CI; a PR that lowers coverage below the floor fails. The floor is raised as the program progresses. This makes the non-complacency rule systemic instead of discretionary.

## Prioritized roadmap (state-safety first)

- [x] **Increment 0 — coverage ratchet** in the `unit` CI job (per-package floor + fail-on-regression) — delivered (#295).
- [x] **P0 — `helpers/` flatten layer (~66 `Flatten*`)**: generic table-driven pattern asserting, for each, nil/omitted API blocks don't panic; emitted keys match the consuming schema; zero-value vs absent is intentional; stable list order; `[]interface{}{}` not nil. (Complements the structural walker #288, which checks schema-fit but not content.) — largely delivered (#296); `helpers` now at 68.1%.
[...]

PR #295, opened 2026-06-14
ci(#293): coverage ratchet for the pure packages (Increment 0)

chrlesur:
## What

Increment 0 of the **S7 test-coverage** program (#293): a coverage ratchet that makes "add tests where they are missing" a **systemic CI guard** instead of a discretionary habit.

- `scripts/coverage_ratchet.sh` runs the pure packages (`internal/client`, `internal/provider`, `internal/provider/helpers`) under `-race` with coverage, and **fails if any package drops below its recorded floor**.
- `scripts/coverage_floor.txt` holds the per-package floor, locked at today's baseline. It is raised via `--update` as the suite grows, and **never lowered by hand**.
- The `unit` CI job now calls the script instead of the bare `go test`.

## Baseline floor (locked)

| Package | Floor |
|---|---|
| `internal/client` | 10.8% |
| `internal/provider` | 8.5% |
| `internal/provider/helpers` | 12.9% |

These rise as the S7 increments land (the in-flight helpers batch will push `helpers` well above 12.9%, after which we re-baseline).

## Non-complacency proof (the ratchet bites)

- Gate mode at the real floor → **passes** (exit 0).
- Floor artificially inflated (`helpers` → 99%) → `RATCHET FAIL … 12.9% < floor 99.0%`, **exit 1**. A ratchet that can't fail would be theater; this one does.

## Safety / scope

- **No production code touched** — only a script, a data file, and the CI wiring.
- Portable bash 3.2+ (no associative arrays) so macOS and the ubuntu CI runner behave identically.
- `-race` requires CGO, available on the GitHub ubuntu runners.

## Local verification

`gofmt -l .` empty · `go vet ./...` clean · `staticcheck ./...` 0 findings · ratchet gate green (10.8 / 8.5 / 12.9 vs floor).

Refs #293

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -89,12 +89,15 @@ jobs:
         # dead stores and suspicious constructs that build/vet miss (#264).
         go run honnef.co/go/tools/cmd/staticcheck@2025.1.1 ./...
 
-    - name: Unit tests (platform-independent packages)
+    - name: Unit tests + coverage ratchet (platform-independent packages)
       run: |
+        # Runs the pure packages under -race with coverage and fails if any
+        # package drops below its recorded floor (scripts/coverage_floor.txt).
         # -race future-proofs the pure packages (token cache, httptest servers,
-        # timers); it is a guard, not state-safety coverage. CGO is available on
-        # the GitHub ubuntu runners.
-        go test -race -count=1 -cover ./internal/client ./internal/provider ./internal/provider/helpers
+        # timers); the ratchet makes "tests where they are missing" a systemic
+        # guard, not a discretionary habit (#264, #293). CGO is available on the
+        # GitHub ubuntu runners.
+        ./scripts/coverage_ratchet.sh
 
   generate:
     runs-on: ubuntu-latest
--- a/scripts/coverage_floor.txt
+++ b/scripts/coverage_floor.txt
@@ -1,0 +1,6 @@
+# Per-package statement-coverage floor (percent) for the pure packages.
+# Raised as coverage improves (scripts/coverage_ratchet.sh --update);
+# CI fails if coverage drops below these values. Never lower by hand.
+internal/client 10.8
+internal/provider 8.5
+internal/provider/helpers 12.9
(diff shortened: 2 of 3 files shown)


chrlesur:
## Review — Codex adversarial pass (read-only)

**Reviewer role:** Codex (gpt-5.5 high), independent from the pilot model that authored the change.

**Findings (verbatim, condensed):**

- **C1 Gate reality:** non-vacuous. `awk 'BEGIN{exit !(p+e<f)}'` fails exactly when coverage < floor − 0.05. Real drops pass only within the 0.05 float-noise tolerance. *Soft spot:* an absent floor entry defaulted to 0 → could silently disable the gate. → **Addressed** (commit b600aef): a listed package with no recorded floor now FAILs in gate mode (proven: removing the helpers floor line → exit 1).
- **C2 Correctness:** runs the same 3 pure packages under -race; a failing `go test` still fails CI (direct command under `set -euo pipefail`, stdout not redirected); `-covermode=atomic` correct with -race; coverage parse correct.
- **C3 Portability:** bash 3.2 + bash 5 OK (no associative arrays, no GNU-only flags); `bash -n` passes; word-splitting safe (import paths have no spaces).
- **C4 CI wiring:** script committed `100755`, invoked from repo root; gofmt/vet/staticcheck remain independent; timeout risk higher (3 -race runs) but within the 10-min job.
- **C5 State/security:** no production change; no command injection (fixed package list, quoted args, floor via `awk -v` not eval).
- **C6 Residual:** locking the floor at the real baseline (not 0, not aspirational) is the right call; interacts well with the in-flight S7 helpers PR.

**REVIEW VERDICT: GO** — gate is real for the three packages; the only material residual (silent bypass on a missing floor line) has been hardened in b600aef.

---
**Governance:** assignee, milestone v1.8.0, label S7-test-coverage, board Lot=S7 / Status=Review / Release=v1.8.0 — `pr_gov_check.sh` OK.
**Local:** gofmt clean · vet clean · staticcheck 0 · ratchet gate green · bias proofs (inflated floor → fail, missing floor → fail).

**Merge awaits explicit human GO** per the RC rule. Issue #293 stays In Progress.

PR #296, opened 2026-06-14
test(#293): non-complacent unit tests for the helpers flatten layer (batch 1)

chrlesur:
## Summary

Non-complacent unit-test coverage of the pure / state-facing flatten layer in
`internal/provider/helpers` (the #241/#243 bug class: a datasource read breaks
because a flatten helper emits a key the schema does not declare, panics on a
bare object, drops/leaks data, or returns the wrong shape).

This is **test-only**. No production code is modified.

It complements — and does not duplicate — the existing structural walker
`TestDatasourceFlattenOutputsFitTheirSchemas` (provider package), which feeds a
fully-populated struct and checks the output *fits* the datasource schema. This
PR adds what the walker does not check: nil/zero-input safety, content
correctness, secret preservation, exact key sets, stable list order,
empty-slice-as-non-nil intent, conditional-block emission, and strict parsing.

Coverage of `internal/provider/helpers`: **12.9% → 58.3%**.

## A) Generic invariant harness

`helper_flatten_invariants_test.go` — a reusable, table-driven harness asserting
the generic contract on a **bare (zero-valued) struct pointer** for **53**
registered `Flatten*` helpers:

- **(i)** no panic on a fully zero-valued struct pointer (the realistic Read
  contract);
- **(ii)** the returned map is non-nil (a nil map is a silent state drop);
- **(iii)** every emitted slice is a non-nil empty slice on empty input, unless
  the `(helper, key)` pair is in the explicit `knownNilSliceGaps` registry.

**Design notes (flagged for the reviewer):**

- **Nil-pointer safety is a documented non-goal.** The helpers do not guard a
  nil receiver and *would* panic on one (e.g. `FlattenRole(nil)`). The Read
  paths always decode a non-nil struct, so the harness pins the realistic
  contract (zero-valued struct). Asserting nil-pointer safety would require a
  defensive guard in ~50 production helpers for an input that never occurs —
  out of scope for a test-only PR.
- **`knownNilSliceGaps` mirrors the walker's `datasourceKnownGaps` pattern.**
  ~18 helpers currently pass a raw nil-able field straight through (e.g.
  `token.Roles`, `pool.Hosts`), yielding a typed-nil slice on empty input. This
  is **SDK-tolerated today** (the writer normalizes `nil` and `[]` to the same
  stored `[]interface{}{}`), so it is a tracked code-quality debt, **not** a
  crash class. The registry makes each gap visible and **ratcheted**: a helper
  later fixed to emit `[]` fails on its now-stale gap entry and forces removal.
  Closing these gaps is left to a follow-up batch (no production change here).

## B) Dedicated non-complacent content tests (batch 1)

| File | Function | What is pinned |
|---|---|---|
| `helper_compute_iaas_opensource_pool_test.go` | `FlattenOpenIaaSPool` | exact key set vs the pools schema (#243); nested cpu/memory/type block content; `machine_manager_id` from the nested object; zero sub-objects don't panic |
| `helper_compute_virtual_machine_test.go` | `FlattenVirtualMachine` | `extra_config` is a `map[string]string` not a list (#241); key→value mapping; empty config is an empty map; last-write-wins; conditional storage/boot_options/replication_config blocks; `triggered_alarms` order |
| `helper_iam_pat_test.go` | `FlattenToken` | the create-only secret is never emitted by name or value; exact datasource key set |
| `helper_object_storage_test.go` | `FlattenStorageAccount`, `FlattenBucket` | `access_secret_key` never emitted (refresh would overwrite it empty, #264 Lot D); `access_key_id`/tags preserved; bucket `versioning` stays a string, `retention_period`/`total_objects` keep int64, exact key set |
| `helper_convert_extra_config_test.go` | `ConvertExtraConfigValue` | strict boolean/numeric parsing, errors not swallowed, string passthrough |
| `helper_update_map_items_test.go` | `UpdateNestedMapItems` | per-index overlay (no cross-index corruption), count/order preserved, empty input yields a non-nil empty slice. **Note:** the explicit-`false` swallowing is *characterized as a tripwire*, **not** guarded here — see "Framing" below |

## Mutation-proof table (non-complacency)

For every covered function, the historical failure mode was injected into the
production helper, the matching test was run (confirmed **RED**), then the
production file was restored from git. Production code is **unchanged** in this
PR.

| # | Function | Injected mutation | Test went RED |
|---|---|---|---|
| M1 | `FlattenOpenIaaSPool` | rename `high_availability_enabled` → `ha_enabled` (undeclared key) | yes |
| M2 | `FlattenOpenIaaSPool` | read `machine_manager_id` from `pool.ID` instead of `pool.MachineManager.ID` | yes |
| M3 | `FlattenVirtualMachine` | emit `extra_config` as a list of blocks instead of a map | yes |
| M4 | `FlattenVirtualMachine` | swap `extra_config` key↔value mapping | yes |
| M5 | `FlattenVirtualMachine` | drop the conditional storage guard (always emit) | yes |
| M6 | `FlattenToken` | leak `token.Secret` under a `secret` key | yes |
| M7 | `FlattenStorageAccount` | emit `access_secret_key` (empty) | yes |
| M8 | `FlattenBucket` | drop `total_objects_deleted` (missing key) | yes |
| M9 | `ConvertExtraConfigValue` | default an invalid boolean to `false` (swallow error) | yes |
| M10 | `ConvertExtraConfigValue` | treat the numeric key as a plain string (no parse) | yes |
| M11 | `UpdateNestedMapItems` | use index `0` for every element (cross-index corruption) | yes |
| M12 | `FlattenRole` (generic ii) | return a `nil` map | yes |
| M13 | `FlattenStorageAccount` (generic iii) | return a typed-nil `tags` slice on empty input (non-gap helper) | yes |
| M14 | `FlattenStorageAccount` (generic i) | index `account.Tags[0]` on a zero struct (panic) | yes |

## Local verification (on the rebased tip)

- `gofmt -l .` → empty
- `go vet ./...` → clean
- `go run honnef.co/go/tools/cmd/staticcheck@2025.1.1 ./...` → **0 findings**
- `go test -race -count=1 -cover ./internal/provider/helpers` → **ok**, coverage **58.3%** (was 12.9%)
- `go build ./...` clean; existing walker tests (`TestDatasourceFlattenOutputsFitTheirSchemas`, `TestEveryDatasourceIsCoveredByTheFlattenWalker`, `TestIAMFeatures*`) still green.

## Production refactor

**None.** All changes are in `*_test.go` files. The mutation proofs were applied
and reverted from git; the working tree's production code is identical to
`rc/v1.8.0`.

## Remaining helpers (next batch)

Not yet given a dedicated content test (the generic harness covers the
canonical-signature ones for the zero-input invariants):

- Dedicated content depth still to add for the generic-harness members beyond
  this batch (e.g. `FlattenHost`, `FlattenDatastore*`, `FlattenOpenIaaSHost`,
  `FlattenOpenIaaSTemplate`, `FlattenBackup*`, `FlattenMarketplaceItem`, the
  IAM `FlattenUser`/`FlattenCompany`/`FlattenTenant`).
- Sub-flatteners with a non-canonical signature, **not** in the generic registry
  and **not** yet covered: `FlattenOpenIaaSVirtualDisk` (takes a `vmID`),
  `FlattenOSDiskData`/`FlattenOSDisksData`,
  `FlattenOSNetworkAdapterData`/`FlattenOSNetworkAdaptersData`,
  `FlattenOpenIaaSOSDiskData`/`FlattenOpenIaaSOSDisksData`,
  `FlattenOpenIaaSOSNetworkAdapterData`/`FlattenOpenIaaSOSNetworkAdaptersData`.
- Closing the `knownNilSliceGaps` debt (emit `[]` explicitly) — a follow-up
  that *would* touch production code and so is intentionally deferred.

## Framing — what the `UpdateMapItems` tests are (and are not)

To avoid a misread: the `UpdateMapItems`/`UpdateNestedMapItems` tests are
**characterization tripwires**, not the explicit-`false` guard.

- `UpdateMapItems` uses `GetOk` semantics, which cannot distinguish an explicit
  `false`/`""`/`0` from an absent attribute. Swallowing the explicit zero is the
  **designed** behaviour of this merge helper; a test here cannot "fail on that
  bug" without redesigning production.
- The test (`TestUpdateMapItemsDoesNotRetainExplicitZeroValues`) therefore
  **pins that limitation as a tripwire**: if the swallowing behaviour ever
  changes, it fails and forces a review of every gating site. It explicitly
  states it "documents the CURRENT behavior, it does not endorse it".
- The **actual** explicit-`false` guard lives in the provider package and is
  itself non-complacent: `TestOptionalComputedBooleansHavePatchGuards` is a
  registry test that fails if any Optional+Computed boolean is not raw-config
  gated (or documented), with the gating implemented in
  `openIaasVMDesiredPropertiesFromResourceData`, `osAdapterTxConfigured`,
  `buildVMwareBootOptionsFromRaw`, etc.

So the #246 explicit-`false` class **is** covered by a fail-on-the-bug test —
in the provider package — and this PR deliberately does not duplicate it.

## Independent review

An adversarial Codex pass raised this framing as a NO-GO (reading the
characterization test as a failed guard). The pilot review adjudicated it as a
framing/accounting point, not a code defect, after independently re-injecting
the #241 (`extra_config` as a list) and #243 (undeclared pool key) mutations and
confirming both tests go RED, and after verifying the real explicit-`false`
guard exists and is non-complacent. See the pilot review comment on this PR.

---

**Awaiting explicit human GO — do not merge by automation.**

Refs #293

### DIFF
--- a/internal/provider/helpers/helper_compute_iaas_opensource_pool_test.go
+++ b/internal/provider/helpers/helper_compute_iaas_opensource_pool_test.go
@@ -1,0 +1,139 @@
+package helpers
+
+import (
+	"testing"
+
+	"github.com/cloud-temple/terraform-provider-cloudtemple/internal/client"
+)
+
+// TestFlattenOpenIaaSPoolContent is the dedicated, non-complacent content test
+// for the #243 culprit. FlattenOpenIaaSPool feeds the iaas_opensource_pool(s)
+// datasource, whose schema declares EXACTLY: id, name, label, internal_id,
+// machine_manager_id, high_availability_enabled, master, hosts, memory, cpu,
+// type. If the helper emits any key the schema does not declare, d.Set fails
+// with "Invalid address to set" and the whole datasource read breaks (#243).
+//
+// The structural walker already checks the fully-populated output FITS the
+// schema. This test adds what the walker does NOT:
+//   - the emitted key SET is exactly the declared one (no extra, no missing) —
+//     pinned against the live datasource schema, so a future drift on either
+//     side is caught;
+//   - the nested single-element blocks (memory, cpu, type) carry the right
+//     sub-keys with the right values;
+//   - the machine_manager_id is read from the nested MachineManager.ID;
+//   - a pool with empty sub-objects does not panic and still emits every block.
+func TestFlattenOpenIaaSPoolContent(t *testing.T) {
+	pool := &client.OpenIaasPool{
+		ID:                      "pool-1",
+		InternalID:              "int-1",
+		Name:                    "prod-pool",
+		Label:                   "Production",
+		HighAvailabilityEnabled: true,
+		Master:                  "host-master",
+		Hosts:                   []string{"host-a", "host-b"},
+		MachineManager:          client.BaseObject{ID: "mm-1", Name: "vc-1"},
+	}
+	pool.Memory.Usage = 40
+	pool.Memory.Size = 100
+	pool.Cpu.Cores = 8
+	pool.Cpu.Sockets = 2
+	pool.Type.Key = "xcp-ng"
+	pool.Type.Description = "XCP-ng pool"
+
+	got := FlattenOpenIaaSPool(pool)
+
+	// --- exact key set against the live datasource schema -----------------
+	wantKeys := map[string]bool{
+		"id": true, "name": true, "label": true, "internal_id": true,
+		"machine_manager_id": true, "high_availability_enabled": true,
+		"master": true, "hosts": true, "memory": true, "cpu": true, "type": true,
+	}
+	for k := range got {
+		if !wantKeys[k] {
+			t.Errorf("FlattenOpenIaaSPool emits undeclared key %q; the pool datasource schema would reject it with 'Invalid address to set' (#243)", k)
+		}
+	}
+	for k := range wantKeys {
+		if _, ok := got[k]; !ok {
+			t.Errorf("FlattenOpenIaaSPool is missing the expected key %q", k)
+		}
+	}
+
+	// --- scalar content ----------------------------------------------------
+	assertEq(t, "id", got["id"], "pool-1")
+	assertEq(t, "name", got["name"], "prod-pool")
+	assertEq(t, "label", got["label"], "Production")
+	assertEq(t, "internal_id", got["internal_id"], "int-1")
+	assertEq(t, "high_availability_enabled", got["high_availability_enabled"], true)
+	assertEq(t, "master", got["master"], "host-master")
+	// machine_manager_id is the NESTED MachineManager.ID, not the pool ID.
+	assertEq(t, "machine_manager_id", got["machine_manager_id"], "mm-1")
+
+	hosts, ok := got["hosts"].([]string)
+	if !ok {
+		t.Fatalf("hosts has type %T, want []string", got["hosts"])
+	}
+	if len(hosts) != 2 || hosts[0] != "host-a" || hosts[1] != "host-b" {
+		t.Errorf("hosts = %v, want [host-a host-b] in order", hosts)
+	}
+
+	// --- nested single-element blocks --------------------------------------
+	mem := singleBlock(t, "memory", got["memory"])
+	assertEq(t, "memory.usage", mem["usage"], 40)
+	assertEq(t, "memory.size", mem["size"], 100)
+
+	cpu := singleBlock(t, "cpu", got["cpu"])
+	assertEq(t, "cpu.cores", cpu["cores"], 8)
+	assertEq(t, "cpu.sockets", cpu["sockets"], 2)
+
+	typ := singleBlock(t, "type", got["type"])
+	assertEq(t, "type.key", typ["key"], "xcp-ng")
+	assertEq(t, "type.description", typ["description"], "XCP-ng pool")
+}
+
+// TestFlattenOpenIaaSPoolEmptySubObjectsDoNotPanic proves the helper survives a
+// pool whose nested objects are all zero (a bare API object). The blocks must
+// still be emitted as single-element lists so the schema shape stays stable.
+func TestFlattenOpenIaaSPoolEmptySubObjectsDoNotPanic(t *testing.T) {
+	defer func() {
+		if r := recover(); r != nil {
+			t.Fatalf("FlattenOpenIaaSPool panicked on a zero-valued pool: %v", r)
+		}
+	}()
+
+	got := FlattenOpenIaaSPool(&client.OpenIaasPool{})
+
+	// machine_manager_id comes from a nested struct; on a zero pool it must be
+	// the empty string, NOT a panic.
+	assertEq(t, "machine_manager_id", got["machine_manager_id"], "")
+
+	// The three nested blocks are always single-element lists, even when empty.
+	for _, key := range []string{"memory", "cpu", "type"} {
+		block := singleBlock(t, key, got[key])
+		if len(block) == 0 {
+			t.Errorf("%s block is empty; expected the zero-valued sub-keys to be present", key)
+		}
+	}
+}
+
+// singleBlock asserts a flatten value is a single-element nested block list and
+// returns its only map. Shared by the pool sub-block assertions.
+func singleBlock(t *testing.T, key string, v interface{}) map[string]interface{} {
+	t.Helper()
+	list, ok := v.([]map[string]interface{})
+	if !ok {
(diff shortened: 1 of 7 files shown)


chrlesur:
Awaiting independent reviewer.

This PR is in `Review` (project #11, Lot S7, Release v1.8.0) pending validation by another model or a human. It is **test-only** — no production code changed. The mutation-proof table in the PR body shows each covered function went red under its injected historical failure mode, then was reverted.

Reviewer focus suggestions:
- the two documented design decisions in `helper_flatten_invariants_test.go`: nil-pointer safety as a non-goal, and `knownNilSliceGaps` as a ratcheted, SDK-tolerated debt registry (no production change);
- confirm the secret-non-emission contracts for `FlattenToken` / `FlattenStorageAccount` match the resource/datasource schemas.

Do not merge without explicit human GO.

chrlesur:
## Pilot review (independent) — PR #296

Reviewer role: pilot model, independent re-verification of an implementer-agent deliverable. I did not trust the PR body; I checked the code and re-ran the proofs myself.

### Independently confirmed (strong)
- **Scope is test-only.** `git diff rc/v1.8.0...` = 7 files, +941, all `internal/provider/helpers/*_test.go`. Zero production change. Branch is rebased on current rc (e58c96f).
- **Coverage** re-run by me: `internal/provider/helpers` 12.9% → **58.3%** under `-race`. gofmt / vet / staticcheck@2025.1.1 all clean on the package.
- **Non-complacency re-proven by me** (not taken on trust). I re-injected two mutations into production helpers:
  - `extra_config` emitted as a list instead of `map[string]string` (#241) → `TestFlattenVirtualMachineExtraConfigLastWriteWins` **RED**.
  - undeclared key added to `FlattenOpenIaaSPool` (#243) → `TestFlattenOpenIaaSPoolContent` **RED** ("emits undeclared key … 'Invalid address to set' (#243)"). Restored → green.
  So the #241/#243/secret content tests genuinely fail on the historical bug.

### Adversarial pass (Codex) — verdict and adjudication
Codex returned **REVIEW VERDICT: NO-GO**, single reason: the `UpdateMapItems`/`UpdateNestedMapItems` tests pass *with* the historical explicit-false swallowing behaviour instead of failing on it.

Pilot adjudication: **this is a framing/accounting issue, not a code defect.**
- The tests are honestly named (`TestUpdateMapItemsDoesNotRetainExplicitZeroValues`) and documented as **pinning a known limitation** of the `GetOk`-based merge (#246 class): "documents the CURRENT behavior, it does not endorse it." They are **characterization tripwires**, not claimed guards — if the swallowing behaviour ever changes, they fail and force a review of every gating site.
- Swallowing explicit-false is the **designed** behaviour of a `GetOk` merge; you cannot make this test "fail on the bug" without redesigning production. The real protection is architectural.
- The **actual** explicit-false guard exists and is itself non-complacent: `TestOptionalComputedBooleansHavePatchGuards` (provider package) is a registry test that **errors** if any Optional+Computed boolean is not raw-config-gated (or documented), with the gating implemented in `openIaasVMDesiredPropertiesFromResourceData`, `osAdapterTxConfigured`, `buildVMwareBootOptionsFromRaw`, etc. I verified it.

So the #246 bug class **is** covered by a fail-on-the-bug test — in the provider package — and #296 correctly does not duplicate it. The only correction: the PR summary must **not** present the `UpdateMapItems` tests as the explicit-false non-complacency proof; they are characterization tripwires, with the guard tracked by `TestOptionalComputedBooleansHavePatchGuards`. The test code already states this; the PR summary is being aligned.

### Minor (non-blocking)
- Generic harness registers ~55 helpers (PR says 53) — immaterial count.
- `knownNilSliceGaps` leaves some list fields (`roles`, `hosts`, `permissions`, `categories`) able to emit typed-nil on empty input — accepted, visible debt that ratchets both ways (a fixed helper fails until its gap entry is removed), mirroring `datasourceKnownGaps`.

### Conclusion
Substantively sound, test-only, genuinely non-complacent for #241/#243/secret leakage (re-proven). The NO-GO is resolved as a framing correction, not a code change. Recommending GO once the PR summary framing is aligned; **merge awaits explicit human GO** per the RC rule. Issue #293 stays In Progress.

chrlesur:
READY FOR RE-REVIEW — PR body reworded: the `UpdateMapItems`/`UpdateNestedMapItems` tests are now explicitly framed as characterization tripwires (not the explicit-`false` guard), with the real guard named (`TestOptionalComputedBooleansHavePatchGuards`, verified non-complacent). No code change; test-only diff unchanged. The Codex NO-GO is adjudicated as a framing point — see the pilot review above.

Issue #302, opened 2026-06-14
Audit: 403-as-not-found convention across 51 client reads (403 is Forbidden, not not-found)


chrlesur:
## Summary

The provider's lenient Read path uses `requireNotFoundOrOK(resp, 403)` at **51 client call sites**, which treats a `403 Forbidden` as "not found". HTTP-semantically that is wrong: `403` = Forbidden (a permission/authorization problem), not `404` = absent. This issue tracks auditing and (safely) correcting that convention.

## What was verified (state-safety is NOT compromised)

A transient/permission `403` does **not** silently drop a managed resource from state. Every state-facing resource Read confirms deletion with an independent **strict** read (`requireOK`, where `403` is an error) before any `d.SetId("")`, and explicitly refuses to drop on an access-denied/ambiguous answer (deletion-evidence doctrine, #275/#271/#273/#281). Verified on the OpenIaaS VM/disk/NIC/replication-policy, VMware, object-storage and PAT resources.

So the impact of the convention is, today:
- **Datasource reads** (the majority of the 51 sites): a `403` surfaces as "not found" instead of "access denied" — a **misleading diagnostic**, read-only, no state harm.
- **Resource reads**: the lenient `403→not-found` is a deliberate fast-path tolerance, backed by the strict-read deletion-evidence guard — **no silent state drop**.

## The open question (why we must NOT fix this blind)

The convention exists because some (older) Cloud Temple services appear to return `403` for an **absent** resource. If that is true per service, then passing `404` instead would make a genuinely-absent resource raise "access denied" — breaking not-found handling. The **new VPC API cleanly separates 403/404** (and was fixed to pass `404` in #301). For the older services we do **not** know the real behaviour, and swaggers here have been unreliable (e.g. IAM `subFeatures`).

## Plan (sequenced, not blind)

1. **Determine the real per-service behaviour**: does each API return `403` or `404` for an absent resource? Best done empirically via the recette acceptance harness (#300) on the dedicated tenant, and/or an API-team question.
2. Where a service returns `404` for absent: pass `404` so a `403` surfaces as a clear access-denied error (semantic fix, safe — same change as #301).
3. Where a service genuinely returns `403` for absent: keep the tolerance but document it, and consider improving the diagnostic.
4. Keep the deletion-evidence guard as the state-safety backstop regardless.

## Scope / risk

[...]

PR #301, opened 2026-06-14
feat(#238): VPC datasources (Phase 1, read-only)

chrlesur:
## Summary

Phase 1 of the VPC provider surface (issue #238): the **read-only, additive** VPC datasources. No write resource, no change to any existing resource/datasource schema. Targets `rc/v1.8.0`.

## Datasources added (8)

| Datasource | Endpoint | Notes |
| --- | --- | --- |
| `cloudtemple_vpc_vpc` / `vpcs` | `GET /vpc/v1/vpc[/{id}]` | |
| `cloudtemple_vpc_private_network` / `private_networks` | `GET /vpc/v1/private_networks[?vpcId][/{id}]` | `ip_address` is the CIDR; `?vpc_id` filter on the plural |
| `cloudtemple_vpc_static_ip` / `static_ips` | `GET /vpc/v1/static_ips/{id}`, `/mac/{mac}`, `GET /vpc/v1/private_networks/{id}/static_ips[?virtualMachineId]` | singular: `ExactlyOneOf(id, mac_address)`; plural: `private_network_id` + `?virtual_machine_id` |
| `cloudtemple_vpc_floating_ip` / `floating_ips` | `GET /vpc/v1/floating_ips[?vpcId][/{id}]` | `?vpc_id` filter on the plural |

All require the `vpc_read` role. Nested API objects are flattened to flat scalar attributes (`vpc_id`, `floating_ip_id`/`floating_ip_address`, `static_ip_id`/`static_ip_address`, `virtual_machine_id`, `network_adapter_id`, …); nullable associations collapse to empty strings so every Computed attribute is always set.

## Swagger validation (vs `swagger.vpc.yml` v0.15.0) — and #239 fixes

#239 was used as a reference and validated field-by-field against the swagger. Fixes/decisions:

- **`VPC.internetIp` is `nullable`** — #239 modelled it as a plain `string`. Changed to `*string` so a null decodes to nil (not `""` conflated with a real value); the flatten still emits `""`.
- **Nested floating-IP address field name** — the swagger uses **`floatingIp.ipAddress`** on a `StaticIp` but **`staticIp.address`** on a `FloatingIp`. Modelled with two distinct named types (`StaticIPFloatingIP{ipAddress}`, `FloatingIPStaticIP{address}`) and pinned by a client decode test + a flatten test (mutation-proven).
- **Read-only scope** — dropped #239's floating-IP `Create/Update/Bind/Unbind/Delete` write methods (Phase 3).
[...]

PR #303, opened 2026-06-14
fix(#238): treat VPC 403 as not-found (live API returns 403 for absent)

chrlesur:
## What

Reverts the `403→404` change made in #301 for the five VPC single-reads: they pass `requireNotFoundOrOK(resp, 403)` again, so a `403` is treated as not-found.

## Why — live evidence (the recette tenant earned its keep)

#301 changed the VPC reads to `404` based on the **swagger**, which documents `403` (Forbidden) and `404` (not found) as distinct. A read-only probe against a **real tenant** proved the swagger wrong:

- `VPC.Read(<absent id>)` and `PrivateNetwork.Read(<absent id>)` → **`403 Forbidden`**. The live VPC API returns **403 for an ABSENT resource**, not 404.
- The datasources themselves read real data correctly (1 VPC + 4 private networks with their CIDRs).

So treating `403` as not-found is the **correct** behaviour and matches the provider-wide convention (51 call sites): a lookup of an absent VPC/network now returns a clean not-found instead of a confusing `access denied` error.

## Trade-off (API limitation, not a provider bug)

The API conflates absent and forbidden (both → `403`, an anti-enumeration pattern), so a genuine permission denial is indistinguishable from absence. The provider cannot resolve this. The **swagger incorrectly documents 404** for not-found — to be reported to the API team (tracked in #302).

## Verification

- Unit test flipped to pin `403 → (nil,nil)` with the live-evidence rationale; green.
- Re-validated **live**: `Read(absent)` now returns a clean not-found (no error).
- gofmt / vet / staticcheck clean; `go test -race ./internal/client` green.

Refs #238

### DIFF
--- a/internal/client/vpc_floating_ip.go
+++ b/internal/client/vpc_floating_ip.go
@@ -56,7 +56,7 @@ func (f *VPCFloatingIPClient) List(ctx context.Context, filter *FloatingIPFilter
 }
 
 // Read retrieves a single floating IP by ID. It returns (nil, nil) when the
-// floating IP does not exist (404).
+// floating IP does not exist (403; the API returns 403 for an absent resource).
 func (f *VPCFloatingIPClient) Read(ctx context.Context, id string) (*FloatingIP, error) {
 	r := f.c.newRequest("GET", "/vpc/v1/floating_ips/%s", id)
 	resp, err := f.c.doRequest(ctx, r)
--- a/internal/client/vpc_floating_ip.go
+++ b/internal/client/vpc_floating_ip.go
@@ -64,7 +64,7 @@ func (f *VPCFloatingIPClient) Read(ctx context.Context, id string) (*FloatingIP,
 		return nil, err
 	}
 	defer closeResponseBody(resp)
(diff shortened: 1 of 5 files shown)


chrlesur:
## Pilot + Codex review

**Live-evidence driven.** A read-only probe against a real tenant (the agentic VPC `FW_AGENTIC`) showed:
- the VPC datasources read real data correctly (1 VPC + 4 private networks with CIDRs);
- `Read(<absent id>)` returns **403 Forbidden**, not 404 — the swagger is wrong.

So this reverts #301's `403→404` to `403→not-found`, matching the live API and the provider-wide convention. Re-validated **live**: `Read(absent)` now returns a clean not-found (no error). The unit test is flipped to pin `403 → (nil,nil)` with the live rationale; comments corrected to 403.

**Codex review: REVIEW VERDICT: GO** (all 5 sites reverted, test non-complacent, scope = 5 client files + the test only). Pilot authored the change → independent validation = the Codex GO + the live probe + the human GO before merge.

CI below. **Merge awaits explicit human GO.**

Issue #300, opened 2026-06-14
Functional acceptance harness: full build/destroy of a complete infra on a dedicated recette tenant


chrlesur:
## Summary

A dedicated, **empty** acceptance ("recette") tenant is now available for Terraform functional testing. This enables true end-to-end verification — build a complete infrastructure from nothing, verify it, then destroy it back to empty — which closes the **live-API side** of state safety that the code-side gates (unit tests, schema golden gate #298, schema-vs-flatten walker #288, coverage ratchet) cannot prove on their own. Companion to #8 (integration suite in CI).

Credentials live in an **untracked, gitignored** `.env.recette` (never committed, never printed).

## Why this is different from the current acceptance suite

The existing `internal/provider/tests` suite (driven by `.env.test.dist`, ~83 variables) **points at pre-existing objects** in a populated tenant. The recette tenant is **empty**, so this is a **full-lifecycle** harness: `apply` a complete stack from scratch → assert (plan converges / no drift, reads correct, secrets preserved on refresh, no unrequested PATCH) → `destroy` → assert the tenant is empty again (no orphan resources, no leaked cost).

## What the scoping (cadrage) must deliver

1. **The representative "complete infra"** to build, in dependency order: which resources/datasources to exercise (OpenIaaS VM + network adapter + virtual disk, VPC datasources, object storage, backup assignment, IAM/PAT, tags, …) and a minimal-but-representative topology.
2. **Harness mechanism**: Terraform HCL configs vs Go acceptance tests (`TF_ACC`); the apply → assert → destroy flow; convergence assertions (second `plan` is empty); read-correctness and secret-preservation assertions.
3. **Safety & isolation (top priority)**: a hard guarantee that the harness can target **only the recette tenant, never a production tenant**; `.env.recette` never committed or printed; **reliable teardown** (destroy must always run, even on failure, to avoid orphan resources and cost); idempotency; handling of async activities, transient errors, retries and timeouts.
4. **CI wiring strategy (#8)**: how and when to run (on-demand / scheduled / gated — likely not every PR, given cost and runtime), secret management for the tenant credentials, and guaranteed teardown on CI failure.
5. **Cost / quota** considerations and how to bound them.
6. **Phased rollout**: plan-only/dry-run first → supervised build/destroy on the recette tenant (explicit human GO) → CI integration.

## Out of scope (of the cadrage itself)

[...]

PR #305, opened 2026-06-14
test(#300): recette lifecycle harness Phase 1 (PAT + object storage)

chrlesur:
Refs #300

Phase 1 of the live acceptance ("recette") harness: a fail-closed, isolated
test-support increment restricted to **PAT + object storage** on a dedicated,
truly empty recette tenant. **Nothing was run live**: the harness compiles and
the live TestCases skip without `TF_ACC`; no `apply`/`destroy`, no `TF_ACC=1`
run, no `-sweep`, no credential/secret/tenant-UUID was read, printed or
committed.

This is test-support + examples + docs only. No production provider code
(`internal/client`, schemas, helpers) is changed; the schema snapshot, coverage
floor and the existing `tests/provider_test.go` are untouched.

## Scope (Phase 1)

Exactly the resources buildable on an empty tenant with no compute substrate:
`cloudtemple_iam_personal_access_token`, `cloudtemple_object_storage_bucket`,
`cloudtemple_object_storage_storage_account`, `cloudtemple_object_storage_acl_entry`.

- **Compute deferred** on the Q1 substrate question (no OpenIaaS/VMware here).
- **`global_access_key` excluded** (its Delete is a no-op; would break
  destroy-to-empty).

## Safety model

- **Env-var allowlist.** `CLOUDTEMPLE_RECETTE_TENANT_ID` is required at runtime;
  the guard never falls back to the current tenant. No tenant UUID is committed.
- **Un-skippable, fail-closed guard.** A NEW isolated sub-package
  `internal/provider/tests/recette/` has its own `TestMain` that, in live mode
  (`TF_ACC`) or sweep mode (`-sweep`), authenticates and asserts the JWT
  `scope.id` equals the allowlist **before any TestCase or sweeper runs**,
  aborting fatally on mismatch, then delegates to `resource.TestMain(m)` (runs
  sweepers under `-sweep`, otherwise `m.Run()`). The existing unsafe `tests`
  `TestMain` is never linked in.
- **Redacted errors.** A mismatch aborts with a generic message naming only the
  env var; it never prints the expected or actual UUID. The guard comparison is
  a pure function, table-tested and non-complacent (RED if inverted).
- **Default path is safe.** Without `TF_ACC`/`-sweep`, the guard is skipped, the
  pure unit tests run with no creds and no network, the package never hard-exits.
- **Leak bans.** No `TF_LOG=DEBUG` (the transport masks the PAT body but NOT the
  object-storage `secretAccessKey`); no `set -x`, no env echo in `run.sh`.

## Assertions encoded

- **Convergence / no-drift** — SDK empty-second-plan (no `ExpectNonEmptyPlan`).
- **Read-correctness** — `ImportState`/`ImportStateVerify` for bucket and
  storage_account; PAT with `secret_id` in `ImportStateVerifyIgnore`; `acl_entry`
  has no importer, asserted via `Check` (composite id + parts).
- **Secret preservation on refresh** — capture `secret_id` / `access_secret_key`
  after apply, refresh-only plan, assert the EXACT same value is preserved.
- **Destroy + destroy-to-empty** — SDK auto-destroy; plus bucket/account/PAT
  `ListStrict` (200-only) baselines captured before build and asserted to return
  exactly after destroy. `acl_entry` has no strict list -> documented residual,
  verified indirectly via the bucket/account listings.
- **No-unrequested-PATCH on refresh — DOWNGRADED for V1.** No live HTTP recorder:
  raw `TF_LOG=DEBUG` leaks the storage `secretAccessKey`, and even a path-only
  recorder is unsafe/unreliable (request paths carry UUID-like ids a failed
  assertion would print, and the provider's own `configure` authenticates with
  `POST /iam/v2/auth/personal_access_token`, so a blanket "no POST on refresh"
  is a false positive). V1 relies on the SDK convergence guarantee (an empty
  second plan proves no config-driven mutation) and documents the residual: it
  does not prove "zero mutating HTTP call". A redacted recorder is a separate
  later increment.

## Bucket/ACL drift decision

The bucket resource's `Read` re-populates its INLINE `acl_entry` (`Optional`)
field from the live ACL listing. Managing the same ACL both inline and via the
standalone resource would cause convergence drift. The stack therefore uses the
standalone `acl_entry` resource and keeps the bucket config free of any inline
block. The lifecycle test deliberately does NOT set `ExpectNonEmptyPlan`: if the
bucket `Read` still re-populates inline `acl_entry` and the second plan is
non-empty in live mode, that surfaces as a hard failure and is a genuine
provider finding to report on #300 — not papered over here.

## Files

- `internal/provider/tests/recette/` — guard + isolated TestMain, lifecycle
  tests, baselines, sweeper, guard unit tests.
- `examples/recette/objectstorage_pat/` — inert HCL stack + `run.sh`.
- `.env.recette.dist` (names only), `examples/recette/README.md`,
  `.gitignore` (`.env.recette`).

## CI

Out of Phase 1 scope: CI activation (gated `environment: recette` +
`workflow_dispatch` + `if: always()` teardown) is left as a doc note only; no
live trigger is enabled.

### DIFF
--- a/.env.recette.dist
+++ b/.env.recette.dist
@@ -1,0 +1,33 @@
+# Phase 1 recette harness — environment template (VARIABLE NAMES ONLY, NO VALUES).
+#
+# Copy to .env.recette (gitignored), fill the values, and source it before a
+# supervised recette run. NEVER commit .env.recette. NEVER print these values.
+#
+# The Go harness does NOT read this file: credentials must already be in the
+# process environment (source .env.recette, or CI environment secrets).
+
+# --- Provider credentials (the recette PAT) --------------------------------
+CLOUDTEMPLE_HTTP_SCHEME=
+CLOUDTEMPLE_HTTP_ADDR=
+CLOUDTEMPLE_CLIENT_ID=
+CLOUDTEMPLE_SECRET_ID=
+
+# --- Tenant allowlist (THE safety core) ------------------------------------
+# The recette tenant's JWT scope.id (UUID). The guard authenticates and refuses
+# to run unless the authenticated tenant equals this value. NEVER commit a UUID.
+CLOUDTEMPLE_RECETTE_TENANT_ID=
+
+# --- Phase 1 substrate (must pre-exist on the tenant; datasource-only) ------
+# IAM role name granted to the PAT (PAT create fails with no role).
+CLOUDTEMPLE_RECETTE_IAM_ROLE_NAME=
+# Object-storage role name granted by the acl_entry (referenced by name).
+CLOUDTEMPLE_RECETTE_OS_ROLE_NAME=
+
+# --- Acceptance switch ------------------------------------------------------
+# With TF_ACC unset the live TestCases SKIP and nothing touches the API.
+TF_ACC=1
+
+# --- Standalone HCL stack (examples/recette/objectstorage_pat) --------------
+# Terraform reads these TF_VAR_* for the supervised run.sh stack.
+TF_VAR_iam_role_name=
+TF_VAR_object_storage_role_name=
--- a/.gitignore
+++ b/.gitignore
@@ -82,6 +82,7 @@ terraform.rc
 
 .env
 .env.test
+.env.recette
 .vscode/
 
 # Local-only internal audit subtree (kept on workstation, never published to
--- a/examples/recette/README.md
+++ b/examples/recette/README.md
@@ -66,15 +66,14 @@ allowlisted recette tenant — and abort hard otherwise.*
 ## Teardown layering
 
 1. **SDK auto-destroy** per TestCase (primary), even on a failed step.
-2. **Guarded start-of-run cleanup**: the recette `TestMain` calls an explicit,
-   tenant-guarded cleanup at the start of a live run (clean slate). This is **not**
-   `AddTestSweepers` (which only fires under `-sweep`); the cleanup body is shared
-   with the registered sweeper via one function.
-3. **Manual `-sweep` orphan cleanup** (documented below). Sweepers do **not**
-   auto-run; they fire only with the `-sweep` flag, and each re-asserts the tenant
-   guard before deleting anything. Deletes are scoped to the `test-terraform*`
-   name prefix / `created_by=Terraform` tag; PAT deletes are additionally filtered
-   to the principal's own `UserId` + `TenantId`.
+2. **Explicit `-sweep` clean slate / orphan cleanup** (documented below). There is
+   **no** automatic start-of-run cleanup: the recette `TestMain` mutates nothing on
+   its own (it only runs the tenant guard). A clean slate is obtained by running
+   the explicit, destructive `-sweep` **before** a run — never automatically.
+   Sweepers do **not** auto-run; they fire only with the `-sweep` flag, and each
+   re-asserts the tenant guard before deleting anything. Deletes are scoped to the
+   `test-terraform*` name prefix / `created_by=Terraform` tag; PAT deletes are
+   additionally filtered to the principal's own `UserId` + `TenantId`.
 
 ## How to run (each live step needs explicit human GO)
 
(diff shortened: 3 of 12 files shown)


chrlesur:
Awaiting independent reviewer

Phase 1 recette harness is implemented and all local gates are green (gofmt, go vet, staticcheck 2025.1.1, go build). The recette guard unit tests pass and the live TestCases SKIP without TF_ACC. NOTHING ran live (no apply/destroy, no TF_ACC=1 run, no -sweep). This PR is ready for an independent adversarial review before any human merge GO; the live execution on the recette tenant is a separate, later, explicitly-approved step.

chrlesur:
Fix pushed addressing the adversarial-review NO-GO: **"the documented non-mutating guard pre-flight actually runs TestMain live cleanup and can delete recette-scoped resources before apply."**

Commit `9ff87fb` — `fix(#300): make the live guard pre-flight genuinely non-mutating (remove TestMain auto-sweep)`.

### Root cause
The recette `TestMain` ran an automatic, tenant-guarded `sweepRecette` at the start of any live (`TF_ACC=1`) run. So the `run.sh` pre-flight (`-run ^TestRecetteLiveGuardOnly$`), labelled *"mutates nothing"*, and `TestRecetteLiveGuardOnly`'s *"creates and destroys NOTHING"* claim, were both false: entering `TestMain` swept (deleted `test-terraform*` / `created_by=Terraform` objects) before the selected test ran. Tenant-guarded, yes — but the human-facing non-mutating claim was untrue (state-safety / honesty defect).

### Fix (principle of least surprise)
- Removed the `if liveEnabled() && !sweepEnabled() { sweepRecette(ctx) }` auto-sweep block from `internal/provider/tests/recette/main_test.go` (and the now-unused `fmt` import). In live OR sweep mode `TestMain` now does ONLY: `flag.Parse()` → `guardLiveTenant` (auth + tenant assertion, no I/O mutation) + `abortRedacted` on mismatch → `resource.TestMain(m)`. The default no-creds unit path is unchanged.
- The ONLY destructive cleanup is now the EXPLICIT `-sweep` path: `resource.TestMain(m)` fires the registered sweepers, each re-asserting the tenant guard via `sweepRecette` before deleting. `sweepRecette`, the sweeper registration and the guard/redaction logic are unchanged.
- Updated the stale comments/docs so no false claim remains: `main_test.go` and `sweeper_test.go` comments, `examples/recette/README.md` (teardown layering + explicit `-sweep` section), and `examples/recette/objectstorage_pat/run.sh` (pre-flight now truly non-mutating; clean slate documented as an OPTIONAL, clearly-labelled destructive `-sweep` the operator runs explicitly — never auto-run). `run.sh` keeps `set -euo pipefail`, the EXIT trap-destroy, no `set -x`, no env echo, no `TF_LOG=DEBUG`.

### End-to-end proof the pre-flight is now non-mutating
`run.sh` → `TF_ACC=1 go test -run ^TestRecetteLiveGuardOnly$` → `TestMain` runs only `guardLiveTenant` then `resource.TestMain(m)` → without `-sweep`, `m.Run()` runs only `TestRecetteLiveGuardOnly`, which re-asserts the guard and returns. No sweep path is reachable. `sweepRecette` is reachable only under explicit `-sweep`.

### Local gates (green)
- `gofmt -l` / `go vet ./...` / `staticcheck ./...` / `go build ./...` clean.
- `go test ./internal/provider/tests/recette/...`: guard unit tests PASS, live TestCases + `TestRecetteLiveGuardOnly` SKIP without `TF_ACC`, no hard-exit.
- Non-complacency re-proven: inverting the comparison in `assertRecetteTenant` turns `TestRecetteGuardAssertTenant` RED (`mismatch_aborts`, `exact_match_proceeds`); restored to green.
- `bash -n run.sh` OK; `terraform fmt -check` clean.

No production code, schema snapshot, coverage floor, or `provider_test.go` touched. No secret/UUID committed (only the fictitious `1111…`/`2222…` placeholders remain, in the unchanged `guard_test.go`).

Status remains **Review** — awaiting independent reviewer.

PR #312, opened 2026-06-14
feat(#238): add cloudtemple_vpc_static_ip write resource (VPC<->VM association)

chrlesur:
Refs #238

## Scope

One new write resource: `cloudtemple_vpc_static_ip` — the core #238 association. It allocates a VPC static IP on a private network, bound to a VM network-adapter MAC. The VPC `vpc` / `private_network` / `floating_ip_binding` resources are separate later increments and are NOT touched here. The existing `cloudtemple_vpc_static_ip` **datasource** (and `helpers.FlattenStaticIP`) is reused unchanged; resource and datasource share the type name across the two maps, which is fine.

## Sync-create / async-update asymmetry (verified against the swagger)

The static IP write surface is deliberately asymmetric, confirmed against `DESIGN/swagger/swagger.vpc.yml`:

- **Create — SYNCHRONOUS.** `POST /vpc/v1/private_networks/{privateNetworkId}/static_ips` returns **201 with a body** `{success, message, static_ip_id}`. The id is in the BODY, not a `Location` header — this is NOT an activity. A dedicated client `Create` decodes `static_ip_id`; it does NOT call `doRequestAndReturnActivity` (which would read `Location`). A non-201, or a 201 with an empty `static_ip_id`, is an error.
- **Update — ASYNCHRONOUS.** `PATCH /vpc/v1/static_ips/{id}` returns **201 + `Location`** (an activity) → `doRequestAndReturnActivity` + `WaitForCompletion`.
- **Delete — ASYNCHRONOUS.** `DELETE /vpc/v1/static_ips/{id}` returns **201 + `Location`** (an activity) → `WaitForCompletion`. A `404` is treated as idempotent success.

## Mutability decision (ForceNew vs PATCH), justified against the swagger

`UpdateStaticIpPayload` in the swagger contains **exactly** `resourceDescription` and `macAddress` (`minProperties: 1`), and crucially **no `ipAddress`**. The schema modes follow directly:

| attribute | mode | rationale |
|---|---|---|
| `private_network_id` | Required, **ForceNew**, IsUUID | the create endpoint is per-private-network; it cannot be re-parented |
| `mac_address` | Required, MAC-validated, **mutable (PATCH)** | present in `UpdateStaticIpPayload` → updatable, NOT ForceNew |
| `ip_address` | Optional + **Computed** + **ForceNew** | auto-assigned when omitted (Computed avoids a perpetual diff); NOT in `UpdateStaticIpPayload` → not updatable |
| `resource_description` | Optional, **mutable (PATCH)** | present in `UpdateStaticIpPayload` |

The PATCH body is **diff-driven**: only the changed updatable fields are sent. A client test asserts the EXACT PATCH body keys and that `ipAddress` is NEVER present — that pins the mutability fix (see proofs below).

The computed read-back **mirrors the existing datasource exactly** and reuses `FlattenStaticIP` unchanged: IDs only, NO `*_name` fields (`id`, `ip_address`, `mac_address`, `source`, `vpc_id`, `private_network_id`, `virtual_machine_id`, `network_adapter_id`, `resource_description`, `floating_ip_id`, `floating_ip_address`).

## Deletion-evidence design (never-drop + ListStrict)

Following the #275/#281 doctrine and the #303 VPC `requireNotFoundOrOK(resp, 403)` convention (a real access-denied is mapped to nil, i.e. looks like not-found), the resource Read **never `SetId("")` on an inconclusive answer**:

- A nil per-id read is INCONCLUSIVE. The Read confirms via a new client `ListStrict(privateNetworkId)` that accepts **only a complete HTTP 200** (`requireHttpCodes(resp, 200)`); a 206 partial — or any 201/403/5xx — **fails closed** (cannot prove absence).
- listing error / still-present id / missing `private_network_id` → fail closed with an actionable diagnostic, state untouched.
- **Confirmed absent from a complete 200 listing of its own private network** → genuine deletion evidence (the listing is scoped to the resource's own network and provably complete), so the resource is dropped. This is the deliberate, scoped difference from the object-storage/VMware "never drop" helpers, whose listings are unscoped and not provably complete.

Delete is idempotent on 404 and confirms absence via the same strict listing after the activity completes.

## Additive golden diff

`UPDATE_SCHEMA_SNAPSHOT=1 go test ./internal/provider -run TestSchemaSnapshot` regenerated `internal/provider/testdata/schema_snapshot.json`. A key-level semantic diff confirms it is **strictly additive**:

```
[provider_schema] added=[]                            removed=[] changed=[]
[resources]       added=['cloudtemple_vpc_static_ip'] removed=[] changed=[]
[data_sources]    added=[]                            removed=[] changed=[]
```

Exactly one new resource; zero removals, zero retypes; the same-named datasource and every existing entry are byte-for-byte unchanged. This is the likely-safe "new resource" class; no `StateUpgrader` is needed (no existing state shape changes).

## Non-complacency proofs (each RED on the real bug, then restored)

| # | mutation | test that went RED |
|---|---|---|
| 1 | `Create` reads `Location` instead of the body `static_ip_id` | `TestVPCStaticIPCreate/...BODY...` (got `"should-be-ignored"`) + the missing/empty-id cases |
| 2 | add a `ipAddress` field to the PATCH payload | `TestVPCStaticIPUpdate/...forbids ipAddress...` (body contained `ipAddress`) |
| 3 | Read drops (`SetId("")`) on a failing strict listing | `TestReadVPCStaticIPInto/...failing strict listing fails closed` |
| 4 | `ListStrict` also accepts 206 | `TestVPCStaticIPListStrict/Partial_Content_is_rejected...` |
| 5 | drop a flatten key (`vpc_id`) from the schema | `TestVPCStaticIPResourceSchemaKeyset` |

Each was reverted and the suite is green.

## State-safety

The client's Terraform state is the overriding priority. Create sets the id immediately (before the read) so a later failure cannot orphan the just-created static IP. Read never auto-removes a resource on an inconclusive/access-denied/partial answer — it only drops on a complete, scoped 200 listing that confirms absence; every other branch fails closed and preserves the state. The PATCH is diff-driven and never sends the non-updatable `ipAddress`. No secret or tenant UUID is read, printed or committed.

## Gates (all green locally)

`gofmt -l` (clean), `go vet ./...` (clean), `staticcheck ./...` v2025.1.1 (clean), `go build ./...`, `go test -race` on `internal/client`, `internal/provider`, `internal/provider/helpers` (all ok; the `*/tests` acceptance suites need `SHIVA_TEST_*` and are not in the unit gate). Golden additive (above); the flatten walker stays green; coverage ratchet floors raised (client 14.1→16.0, provider 8.7→9.1) and committed; `go generate ./...` produced the resource doc.

## Live validation pending

Live validation on a real tenant is the independent reviewer's job (adversarial Codex DIFF pass + live run). Nothing here was run against a live API.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


chrlesur:
Awaiting independent reviewer

Issue #316, opened 2026-06-15
feat(tooling): cmd/ct-validate — parametrizable endpoint validation & resilience harness


chrlesur:
Related to #300 (recette harness). A standalone, parametrizable Go CLI that exercises the provider's endpoints through the client library and reports **where it squeaks**.

## Why

Today (2026-06-15) the VPC write path showed intermittent API failures (`[ComputeManager] None of the workers were able to respond`, and `502 Bad Gateway`). We need a repeatable way to quantify reliability across **all** endpoints, with realistic business cycles, without hammering a fragile shared API.

## Scope

- Reuses `internal/client` (auth + all 8 services: Activity, Backup, Compute VMware/OpenIaaS, IAM, Marketplace, ObjectStorage, Tag, VPC).
- **Read layer**: sweep the read/list endpoints across every service (safe, broad).
- **Write layer (opt-in)**: complete business cycles with guaranteed teardown — VPC (static IP → bind floating IP → unbind → delete), Object Storage (bucket → ACL → delete), Backup (SLA assign → inventory), IAM PAT (create → read → delete), Compute OpenIaaS VM lifecycle (heavy, gated).

## Parametrization

`-runs N`, `-concurrency P`, `-cycles a,b,c`, `-write` (off by default), `-tenant-env`, `-timeout`, `-json`.

## Safety (mandatory)

- Low default concurrency; write disabled by default.
- **Circuit breaker**: abort immediately when the rolling failure rate exceeds a threshold or after N consecutive failures — never keep hammering a failing API.
- Guaranteed cleanup of every created resource (deferred teardown + final sweep), one delete attempt + bounded transient-retry only.

## Output

Per cycle/endpoint: runs, success %, latency p50/p95, and a failure-category histogram (transient ComputeManager / 502 gateway / 4xx / timeout / other). Table + optional JSON.

PR #330, opened 2026-06-15
ct-validate: confirm VMware compute delete 403-as-absent (no false teardown alarms)

chrlesur:
## Symptom
A fully successful VMware write cycle (`compute_vmware_lifecycle`) — create + explicit deprovision all `200` — still printed 3 `TEARDOWN FAILURES (possible orphans — investigate)` and exited non-zero, with `403 (Forbidden.)` for the adapter, disk and VM. **No real orphan** existed.

## Root cause
The deferred teardown (the always-on backstop) re-deletes what the explicit deprovision already removed. The VMware compute API returns **`403` for an ABSENT resource** (it conflates absent/forbidden — the #303 pattern), but the teardown idempotency only treated **`404`** as "already gone", so a 403-on-absent surfaced as a false failure. Proof it is *absent* not *forbidden*: the explicit delete with the same PAT succeeded moments earlier.

## Fix
`confirmComputeDeleteErr(err, stillExists, id)`:
- `404` → idempotent success;
- `403` → **confirmed via a by-id existence re-check** (`seam.Exists` → `Read` by id, which maps 403/404 → not-found): accepted **only** when proven absent, surfaced when it **still exists**, failed **closed** when the re-check itself errors;
- other errors surface unchanged.

The re-check is **by id and independent of the parent VM**, so it stays valid when the deferred leaf teardown runs *after* the explicit deprovision already deleted the parent VM (a VM-scoped listing would fail there — the round-1 review caught this). Mirrors the VPC `staticIPDeleteErrResult` (#312) confirm-before-accept doctrine. Wired into the VM/disk/adapter deferred teardowns; the explicit deprovision path is unchanged (resources present there → `200`, never a 403-on-absent).

## Tests / review
Mutation-proven: helper decision (404/403-absent/403-present/recheck-error/5xx), wired VM + leaf 403 absent/present, and fail-closed when the re-check errors. Adversarial review (Codex): GO (round 1 NO-GO on parent-scoped confirmation → moved to by-id existence re-check).

Refs #316.

### DIFF
--- a/cmd/ct-validate/cleanup_seams_vmware.go
+++ b/cmd/ct-validate/cleanup_seams_vmware.go
@@ -3,9 +3,46 @@ package main
 import (
 	"context"
 	"fmt"
+	"net/http"
 
 	"github.com/cloud-temple/terraform-provider-cloudtemple/internal/client"
 )
+
+// confirmComputeDeleteErr resolves a VMware compute delete outcome. A 404 is a
+// definitive not-found → success (idempotent). A 403 is AMBIGUOUS: the VMware
+// compute API returns 403 for an ABSENT resource as well as a forbidden one (the
+// #303 conflation) — typically when the explicit deprovision already deleted the
+// resource and the deferred backstop re-deletes it. The 403 is CONFIRMED by
+// re-checking the resource's OWN existence (a Read by id, which maps 403/404 →
+// not-found): accepted ONLY when it is proven absent; surfaced when it still
+// exists; failed CLOSED when existence cannot be determined (a 5xx/transport
+// error on the re-check). Any other delete error surfaces unchanged.
+//
+// The existence re-check is BY ID and independent of the parent VM — so it stays
+// valid even when the deferred leaf teardown runs AFTER the explicit deprovision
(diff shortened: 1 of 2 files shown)


chrlesur:
**Known limitation (documented, not blocking).** This PR's `confirmComputeDeleteErr` confirms a deferred-teardown `403` via a by-id `Exists` re-check (a `Read`), but the VMware client maps both `403` and `404` to "absent". So a *present-but-forbidden* resource (a PAT that can create but neither delete nor read it) would be read as "gone" and a real orphan could be masked.

Why we are not hardening it now:
- It has never been observed — the recette PAT has full rights, so a `403` here means absence.
- If the delete right were missing, the explicit deprovision phase (404-only) would already surface the failure.
- This is the internal `ct-validate` harness: a false "absent" would at worst leave a *test* orphan, not touch a client's Terraform state.

If the present-but-forbidden case is ever observed (false teardown orphans on a healthy run), harden with a same-cycle delete proof at the sweep level (the adapted mirror of the OpenIaaS fix in #334, which replaced the by-id `Exists` with `confirmComputeDeleteByPriorDelete`). Tracked as backlog.

PR #335, opened 2026-06-15
fix(#316): OpenIaaS lifecycle — drop the disk disconnect (delete works attached)

chrlesur:
Refs #316

> **Stacked on #334** (base `s9/ct-validate-openiaas-teardown-403`). GitHub will retarget this PR to `rc/v1.8.0` once #334 merges. Review the per-commit diff (this PR adds one commit on top of Bug B).

## Problem (observed live)

In the live OpenIaaS `compute_lifecycle` run, `compute.openiaas.virtual_disk.disconnect` **failed** (the XOA "Disconnect a virtual disk" activity errored) while the very next `compute.openiaas.virtual_disk.delete` **succeeded** on the still-attached disk.

Root cause (mirror of merged #333, which dropped `Connect`): a disconnect requires a **running** VM (`VM state is halted but should be running`), and this lean lifecycle deliberately leaves the VM **halted**. So the disconnect is superfluous — and worse, it can **block** a delete that otherwise works.

## Fix

Remove the disk disconnect from the harness entirely:

- **Cycle**: drop the explicit `virtual_disk.disconnect` op, its `writeSteps` entry, and its two `r.skip(...)` emissions. The disk block now just deletes.
- **Deferred teardown**: `registerVirtualDiskTeardown` deletes the attached disk **directly** (no `ReadConnections` + `DisconnectAndWait` loop), keeping Bug B's same-cycle `403`-on-absent proof guard (`confirmComputeDeleteByPriorDelete`). `ReadConnections`/`DisconnectAndWait` are removed from the `diskSeam` interface and the seam impl.

For the never-orphan safety net this is also **safer**: a pre-delete disconnect on a halted VM would fail and **block** the delete that works — removing it removes a failure point on the cleanup path. A delete refusal still **surfaces** (never masked): the explicit path is 404-only idempotent, the deferred path accepts a `403` only on same-cycle proof and otherwise fails closed.

## Scope

- The client `VirtualDisk().Disconnect`/`Connect` are **untouched** (used by the provider).
- Mirrors #333 (dropped `Connect`) for the symmetric reason.

## Tests

`go build`, `gofmt -l` (clean), `go vet`, `go test -race ./cmd/ct-validate/...` all green. The disconnect-ordering test is replaced by `TestRegisterVirtualDiskTeardownDeletesDirectlyNoDisconnect` (asserts exactly one op — a direct delete). The `fakeDiskSeam` no longer implements the disconnect methods, so re-introducing a pre-delete disconnect **fails to compile** — a structural guard (compile-time mutation verified).

## Review

PLAN-reviewed adversarially (GO); a multi-lens pre-commit review found no blockers; the committed diff passed an independent adversarial diff review (GO).

### DIFF
--- a/cmd/ct-validate/cleanup_seams.go
+++ b/cmd/ct-validate/cleanup_seams.go
@@ -438,11 +438,13 @@ func (s iamPATSeam) FindIDByName(ctx context.Context, name string) (string, erro
 //
 // Teardown ordering is leaves-first (LIFO over registration order): the network
 // adapter and the user data disk are removed BEFORE the VM anchor, because a VM
-// delete does NOT cascade a user-created disk (it would orphan storage) and a
-// still-connected disk can refuse deletion. Each teardown is registered BEFORE
-// the create it undoes (F3), keyed by a deterministic identity, and finds its
-// resource via a STRICT listing when the create's activity did not resolve the
-// id.
+// delete does NOT cascade a user-created disk (it would orphan storage). The disk
+// is deleted DIRECTLY while still attached — live evidence shows the attached data
+// disk delete succeeds, and an explicit disconnect requires a RUNNING VM (this lean
+// cycle leaves the VM halted), so a pre-delete disconnect would only fail and BLOCK
+// the delete that otherwise works. Each teardown is registered BEFORE the create it
+// undoes (F3), keyed by a deterministic identity, and finds its resource via a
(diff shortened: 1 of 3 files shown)


chrlesur:
Superseded by #336 — this PR was auto-closed by GitHub when its stacked base branch `s9` was deleted on the #334 merge. The fix (same commit) continues in #336, now based on `rc/v1.8.0`.

PR #338, opened 2026-06-15
feat(ct-validate): Terraform validation path — OpenIaaS vm lifecycle (marketplace Ubuntu + disk + start/stop + LAN/static IP/public IP)

chrlesur:
Refs #316

Adds a **real-Terraform** validation path to `ct-validate`, alongside the existing API path. First scenario: the OpenIaaS `vm` full lifecycle, exercised through the actual provider.

## What `scripts/ct-test.sh tf vm` does

Deploy an **Ubuntu VM from the marketplace** → attach a **data disk** → **start → stop → restart**, with a **convergence check** (empty `terraform plan`) after every step, then a **clean destroy**:

```
create+start → plan(empty) → stop → plan(empty) → restart → plan(empty) → destroy
```

```bash
CT_ENV_OPENIAAS=/path/.env.recette-openiaas scripts/ct-test.sh tf vm
```

The convergence check (no permanent drift after apply) is the value of the TF path over the raw API; the API path tests create/read/delete, the TF path adds **convergence** and **teardown** as Terraform sees them.

## Pipeline safety

- **Local provider** via a generated `dev_overrides` CLI config (uses *this* checkout's provider, not a registry release; no `init`).
- **Isolated workdir** (`mktemp -d`): no `terraform.tfstate` ever persists in the repo, so a prior failed run can't contaminate the next (the run-unique `vm_name` guards the remote object; the isolated state guards the local one).
- **Teardown net** (EXIT + INT/TERM traps): always destroys (idempotent; `.destroyed` is marked only on success, so a failed destroy is retried) and removes the workdir — never leaves orphans, even on Ctrl-C.
- **Post-destroy orphan smoke check** (one-directional, FAIL-only): scans the live OpenIaaS listings for the run-named VM and data disk, trusting only a complete HTTP-200 JSON-array listing (auth fail / non-200 / partial → inconclusive, never read as "absent"). It can only FAIL a run (orphan found); it never claims "proven clean" — the destroy (provider delete+wait) is the primary removal proof.

## Config (`examples/ct-test/vm/`)

Substrate (availability zone, storage repository, network, backup policy) is **discovered dynamically** (no hard-coded tenant ids), mirroring the API cycle. Only the Ubuntu marketplace image name is a tunable variable (catalogs differ); cpu/memory are tunable for the image's minimums. `terraform validate` passes against the locally-built provider.

## Scope / follow-ups

- Live `apply`/`destroy` runs create real recette resources — gated behind a human.
- A rigorous independent absence proof (the Go client's `ListStrict` by id, incl. the VM's OS-disk / network-adapter sub-objects) is a documented follow-up; the shell smoke check covers the run-named VM + data disk.
- Other scenarios (vpc, storage, vm-vmware) will get their `examples/ct-test/<scenario>/` next.

## Review

Implemented against the verified provider schemas; the pipeline passed an independent adversarial diff review across 6 rounds (state isolation, teardown net, convergence, verifier rigor, data-disk matching), ending GO. `terraform validate` / `bash -n` / `terraform fmt -check` all clean.

### DIFF
--- a/examples/ct-test/vm/README.md
+++ b/examples/ct-test/vm/README.md
@@ -26,16 +26,43 @@ destroy always runs, even on a mid-lifecycle failure, to never leave orphans.
 
 - **Marketplace deploy** (`marketplace_item_id` + `storage_repository_id`).
 - **Data disk** attached to the VM (`cloudtemple_compute_iaas_opensource_virtual_disk`).
+- **LAN + static IP**: the adapter joins the LAN (an OpenIaaS network) and a managed
+  **static IP** (`cloudtemple_vpc_static_ip`) is allocated on the LAN's VPC private
+  network, bound to the adapter's MAC (address auto-allocated by the API).
+- **Public IP**: a pre-provisioned free floating IP is bound to the static IP
+  (`cloudtemple_vpc_floating_ip_binding`) — the VM is reachable from outside for tests.
 - **Power management** via `power_state` (`on`/`off`) — the stop/start cycle.
 - **Convergence**: no permanent drift after apply (the value of the TF path vs the API).
 - **Clean teardown**: destroy leaves nothing behind.
 
+`terraform output` exposes `lan_static_ip` and `public_ip`.
+
 ## Substrate discovery (no hard-coded ids)
 
-The availability zone, storage repository, network and backup policy are discovered
+The availability zone, storage repository and backup policy are discovered
 dynamically (first usable of each), so the config runs on any OpenIaaS tenant —
 mirroring the API cycle's read-then-pick approach. A backup SLA policy is required
 to power a VM on; any policy in the tenant satisfies it.
+
+**The LAN** is selected **by name** (`var.lan_network_name`) on two correlated
+layers — the OpenIaaS network (where the adapter connects) and the VPC private
+network (where the static IP is allocated), which share the LAN's name by
+convention. Resource `precondition`s assert the name matches **exactly one** network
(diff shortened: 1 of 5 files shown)


chrlesur:
**Update — networking added to the `vm` scenario** (2nd commit).

The VM now joins the **LAN**, gets a managed **static IP**, and a **public (floating) IP** is attached for external reachability tests:

- LAN selected **by name** (`var.lan_network_name`) on both correlated layers (OpenIaaS network for the adapter + VPC private network for the static IP), with `length(...) == 1` **preconditions** that fail closed at plan time on a 0/>1 match (Terraform's `one([])` silently returns `null` — preconditions avoid the silent mis-wire).
- `cloudtemple_vpc_static_ip` bound to the VM adapter's MAC (`ip_address` auto-allocated); the static IP **depends on the VM**, so destroy removes it first (no orphan).
- `cloudtemple_vpc_floating_ip_binding` binds the **first free** floating IP (`static_ip_id == ""`); a precondition fails closed if none is free. Binding only binds/unbinds — never creates/destroys the public IP.
- Outputs `lan_static_ip`, `public_ip`.

`terraform validate` / `fmt` clean. Independent adversarial diff review: one blocking issue caught (`one([])` not fail-closed on 0-match) and fixed with explicit preconditions → **GO**.

Tenant-specific assumptions (documented): the OpenIaaS network and VPC private network share the LAN name; a free floating IP exists. To validate on first live run; adjust `-var lan_network_name=...` if needed.

Issue #340, opened 2026-06-16
client: JWT() returns (nil, nil) when reading the auth response body fails — nil-token panic


chrlesur:
## Symptom

`client.JWT()` can return `(nil, nil)` — a nil token with no error — so a caller that dereferences the token panics.

## Detail

In `internal/client/api.go` (`JWT`, ~lines 244-247), after the auth response is received the body is read with `io.ReadAll`; on a read error the function returns `nil, nil`:

```go
bytes, err := io.ReadAll(resp.Body)
if err != nil {
    return nil, nil   // returns a nil token AND no error
}
```

The caller `doRequest` (~line 285) does `token, err := c.JWT(ctx)`; with `err == nil` it proceeds to `c.doRequestWithToken(ctx, r, token.Raw)`, dereferencing a nil `*jwt.Token` → nil-pointer panic.

## Fix

Return the actual error (`return nil, err`) instead of swallowing it, so a failed auth-response-body read surfaces as a clear error and fails closed.

## Context

Found during the client read-resilience review (Refs #316). Latent: only triggers if reading the auth response body fails.

Issue #339, opened 2026-06-16
client: reads lack a per-request timeout and transient-error retry — a slow/flaky API hangs or fails plan/apply


chrlesur:
## Symptom

Two related failure modes when Terraform talks to a slow or intermittently-failing API:

1. **Hang** — the provider's HTTP client is built with no request `Timeout`, and datasource reads pass Terraform's context through without a local deadline. A slow/stuck endpoint makes a read hang with no client-side bound. Observed live: a `GET /backup/v1/open_iaas/policies` hung ~3 minutes during a datasource read before manual abort.

2. **Plan/apply fails on a single transient 502/5xx** — datasource and resource reads issue a single HTTP call with no retry. One transient upstream `502` (or other `5xx`) on a read fails the whole `plan`/`apply` immediately.

## Root cause

`internal/client/api.go`: `NewClient` builds `&http.Client{Transport: ...}` with no `Timeout` (~line 105). `doRequest`/`doRequestWithToken` (~284-330) issue a single `Do(req)`; only `ctx` bounds the call. `isTransientAPIError` + `go-retry` already exist but are wired only to async activity/backup waiters (e.g. `activity.go`, `backup_job.go`), never to plain reads.

Diagnostic detail: when the same API is exercised by the internal `ct-validate` harness (same Go client) it reports ~99.6–100% OK over 500–1000 calls, but the API is slow (P50 ~4–8s, P95 ~10s) and emits rare intermittent `502` (~0.4%). The harness tolerates that rate via its own run-loop + circuit breaker; the provider has no equivalent read resilience, so the same rare 502 deterministically breaks a `plan`, and a slow endpoint hangs unbounded.

## Fix (agreed)

Add read resilience to the client:

- **Configurable per-request HTTP timeout** — env `CLOUDTEMPLE_HTTP_TIMEOUT` (seconds), default **240s**, applied on the `http.Client` built in `NewClient` (default also applied for callers building a bare `Config{}`). Converts an unbounded hang into a fast, clear failure.
[...]

PR #341, opened 2026-06-16
fix(#339): client read resilience — per-request HTTP timeout + bounded retry on transient read errors

chrlesur:
## Refs #339

A slow or intermittently-failing API made the provider **hang** (no per-request timeout) or **fail a whole `plan`/`apply` on a single transient upstream `502`** (no retry on reads). When the same API is exercised by the internal `ct-validate` harness (same Go client) it reports ~99.6–100% OK because the harness retries; the provider had no equivalent read resilience, so a rare intermittent `502` deterministically broke a plan and a slow endpoint hung unbounded (a `GET /backup/v1/open_iaas/policies` was observed hanging ~3 min).

## Change (`internal/client`)

- **Per-request timeout** — `NewClient` builds the `http.Client` with a `Timeout`, default **240s**, overridable via `CLOUDTEMPLE_HTTP_TIMEOUT` (seconds; fail-safe parse keeps the default on a bad value). Applied to every client it builds **and** to a caller-supplied `HttpClient` that has no timeout of its own, so the hang guard is universal. Converts an unbounded hang into a fast, clear failure.
- **Bounded retry on idempotent GET reads** — a dedicated `doWithRetry` helper (not the shared request path) retries `429`/`5xx` and non-timeout transport errors, draining failed response bodies before each retry. Defaults: 3 attempts, capped-exponential backoff, `Retry-After` honoured on `429`. It **never** retries a configured request timeout, a context error, or a `4xx` (incl. `404`) — those fail closed. **Writes (non-GET) are sent exactly once** (a write may already have started an async activity).
- **Auth POST retry** — retried narrowly inside `JWT` (under the existing mutex, so no re-auth storm; request body rebuilt each attempt), so a transient auth blip does not break every read.
- **Uniform timeout doctrine** — `isTransientAPIError` (the shared classifier used by the activity and backup-job waiters) now **excludes configured timeouts**, so a timed-out poll is no longer retried into a multi-minute stall; `BackupJob.WaitForCompletion` now gates its read-error retry on it (was: retry every error).

Retry is enabled only for a `DefaultConfig`-derived (production) client; a bare `Config{}` keeps single-shot behaviour, so existing unit tests are unchanged.

## Tests (mutation-proven)

`GET 502→200` succeeds; persistent `5xx` is bounded to the attempt budget; a configured timeout and `4xx` are never retried; the auth retry replays the credentials body (not an empty/consumed body); the `http.Client` timeout fires fast; the activity waiter does not retry a timeout. Mutations exercised and verified RED-without-fix: removing the auth body reset; classifying timeouts as retryable (both `doWithRetry` and `isTransientAPIError`).

Gates: `gofmt`, `go vet`, `staticcheck` (clean), `go test ./internal/client -race`, coverage ratchet (**internal/client 18.5% → 21.0%**), `make build`.

## Scope / follow-ups

- The `JWT()` `(nil, nil)`-on-read-error latent nil-pointer is tracked separately in #340 (intentionally not touched here).
- Backoff has no jitter yet (acceptable for 3 attempts; a possible hardening for synchronized retry bursts under high parallelism).
- Pre-existing/adjacent: the provider network-adapter connect/disconnect loops carry a "retry until timeout" comment but return raw errors (so `go-retry` does not actually retry them) — a misleading comment, not in scope here.

### DIFF
--- a/internal/client/activity_unit_test.go
+++ b/internal/client/activity_unit_test.go
@@ -217,6 +217,21 @@ func TestIsTransientAPIError(t *testing.T) {
 	if !isTransientAPIError(fmt.Errorf("wrapped: %w", StatusError{Code: http.StatusBadGateway})) {
 		t.Fatal("wrapped transient StatusError must stay transient")
 	}
+	if !isTransientAPIError(&url.Error{Op: "Get", URL: "https://shiva.example", Err: errors.New("connection reset")}) {
+		t.Fatal("a non-timeout transport error (connection reset) must stay transient")
+	}
+	// A CONFIGURED request timeout (or ctx deadline/cancel) must NOT be transient:
+	// retrying it in a waiter would multiply the per-request bound into a
+	// multi-minute stall (issue #339).
+	if isTransientAPIError(&url.Error{Op: "Get", URL: "https://shiva.example", Err: timeoutErr{}}) {
+		t.Fatal("a configured request timeout must NOT be transient")
+	}
+	if isTransientAPIError(context.DeadlineExceeded) {
+		t.Fatal("a context deadline must not be transient")
+	}
+	if isTransientAPIError(context.Canceled) {
+		t.Fatal("a context cancellation must not be transient")
+	}
 }
 
 func TestIsTransientActivityFailure(t *testing.T) {
--- a/internal/client/api.go
+++ b/internal/client/api.go
@@ -26,6 +26,27 @@ const (
 	HTTPSchemeEnvName       = "CLOUDTEMPLE_HTTP_SCHEME"
 	HTTPClientIDEnvName     = "CLOUDTEMPLE_CLIENT_ID"
 	HTTPClientSecretEnvName = "CLOUDTEMPLE_SECRET_ID"
+	HTTPTimeoutEnvName      = "CLOUDTEMPLE_HTTP_TIMEOUT"
+)
+
+const (
+	// defaultHTTPTimeout bounds each HTTP request (dial + headers + body read).
+	// It is a safety guard: without it a slow/stuck endpoint hangs a read with no
+	// client-side bound. Generous on purpose so it never regresses a legitimately
+	// slow request, while still cutting an unbounded hang. Override with
+	// CLOUDTEMPLE_HTTP_TIMEOUT (seconds).
+	defaultHTTPTimeout = 240 * time.Second
+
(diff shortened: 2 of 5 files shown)


chrlesur:
### Adversarial review trace

Pilot model: Claude. Independent reviewer model: Codex (gpt-5.5, read-only).

- **PLAN review** (before coding) — GO on the fix axis (read retry + configurable timeout), with binding constraints: retry idempotent GET only; not in the shared request path; drain failed bodies; do not retry a configured timeout; rebuild the auth body; mutation-proof tests.
- **DIFF review, round 1 — NO-GO** (founded). The new 120s timeout, combined with the existing waiters that retried on `isTransientAPIError` (which treated any `url.Error`, incl. timeouts, as transient), could turn a single timed-out poll into a multi-minute stall (`Activity.WaitForCompletion` ~9×; `BackupJob.WaitForCompletion` retried every error unconditionally).
- **Remediation** — `isTransientAPIError` now excludes configured timeouts (shared classifier), `BackupJob` gates its read-error retry on it; `NewClient` also guards a caller-supplied timeout-less `HttpClient`. Mutation C added (reverting the classifier makes the activity-waiter test go RED at 9 calls).
- **DIFF review, round 2 — GO**, no blocking findings. One non-blocking note: a pre-existing/adjacent misleading comment on the provider connect/disconnect loops (out of scope here).

Gates green: `gofmt`, `go vet`, `staticcheck`, `go test ./internal/client -race`, coverage ratchet (internal/client 18.5% → 21.0%), `make build`.

`TF prod breaker = RED`: the change broadly affects client HTTP behaviour. The retry net reduces transient failures, but the 120s default timeout could, in a pathological case, fail a single request that legitimately took >120s and previously succeeded. Generous default + `CLOUDTEMPLE_HTTP_TIMEOUT` override mitigate it; to be assessed at RC validation.

chrlesur:
Update: the default per-request timeout is set to **240s** (was 120s) per maintainer decision — still overridable via `CLOUDTEMPLE_HTTP_TIMEOUT`. Rationale: with the GET retry carrying transient-error resilience, the timeout acts as the ultimate anti-(infinite-)hang backstop, so a more generous default further reduces any risk of cutting a legitimately slow request. Note: it is above the ~180s episode observed on `backup/policies`, so for that case the retry/backstop — not the timeout — is what bounds it. `go test ./internal/client -race` and the coverage ratchet stay green (the tests assert against the constant).

Issue #342, opened 2026-06-16
client: JWT cache-hit reads exp with unguarded type assertions — panics on a valid token without a numeric exp


chrlesur:
## Symptom

`JWT()`'s cache-hit branch reads the cached token's expiry with two **unguarded** type assertions:

```go
if c.SavedToken != nil {
    expireAt := c.SavedToken.Claims.(jwt.MapClaims)["exp"].(float64)
    ...
}
```

A syntactically valid JWT that lacks an `exp` claim (or whose `exp` is not a JSON number) is cached on the first call, then **panics** on the next call: `["exp"]` is `nil` and `nil.(float64)` panics (or `Claims.(jwt.MapClaims)` panics if `Claims` is nil).

## Context

Found during the #340 review. #340 makes `JWT` fail closed and stops caching a token on a read/parse **error**, so only successfully-parsed tokens are cached — but a *successfully-parsed token without a usable `exp`* still reaches this branch.

## Fix

Use comma-ok assertions and treat a missing/invalid `exp` as "token unusable / needs refresh" (re-authenticate) rather than panicking:

```go
claims, ok := c.SavedToken.Claims.(jwt.MapClaims)
if ok {
    if exp, ok := claims["exp"].(float64); ok {
        // use exp
    }
}
```

Latent: the production auth token always carries a numeric `exp`, so this is hardening, not an observed crash.

PR #343, opened 2026-06-16
fix(#340): JWT — fail closed on auth read/parse error, never cache a bad token

chrlesur:
## Refs #340

`JWT()` mishandled two auth-response error paths, each leading to a panic:

1. **Read error swallowed** — `io.ReadAll` of the auth response body returning an error caused `JWT` to `return nil, nil` (a nil token **with no error**). The caller `doRequestOnce` then dereferenced `token.Raw` → nil-pointer panic.
2. **Bad token cached on parse error** — `c.SavedToken = token` ran unconditionally after `ParseUnverified`, so a malformed 3-segment token (non-nil token + parse error) was cached; the next cache hit read its claims (`["exp"].(float64)`) and panicked.

## Change (`internal/client/api.go`)

- The read error and the parse error are now **returned** (fail closed, with context).
- **No token is cached on error** — only a successfully-parsed token reaches `c.SavedToken`.
- A transient/garbled auth response now surfaces a clear error and lets the next call re-authenticate, instead of crashing.
- Minor: the local `bytes` was renamed `body` (it shadowed the `bytes` package).

## Tests (mutation-proven)

- A body that errors on read → `JWT` returns an error (not `nil, nil`), no panic, no cached token. (RED if `return nil, nil` is restored.)
- A malformed 3-segment token `aaa.bbb.ccc` → `JWT` returns an error, caches nothing, and a second call does not panic. (RED if the unconditional cache is restored.)

Gates: `gofmt`, `go vet`, `staticcheck` (clean), `go test ./internal/client -race`, coverage ratchet, `make build`.

## Follow-up

The cache-hit branch still reads `exp` with unguarded type assertions, so a syntactically valid token lacking a numeric `exp` would panic on the next call. Out of scope here (distinct from the error paths fixed above); tracked in #342.

### DIFF
--- a/internal/client/api.go
+++ b/internal/client/api.go
@@ -318,15 +318,22 @@ func (c *Client) JWT(ctx context.Context) (*jwt.Token, error) {
 		return nil, err
 	}
 
-	bytes, err := io.ReadAll(resp.Body)
+	body, err := io.ReadAll(resp.Body)
 	if err != nil {
-		return nil, nil
+		// Surface the read failure (fail closed) instead of returning a nil token
+		// with no error — the caller would dereference token.Raw and panic.
+		return nil, fmt.Errorf("failed to read authentication response: %w", err)
 	}
 
-	token, _, err := new(jwt.Parser).ParseUnverified(string(bytes), jwt.MapClaims{})
+	token, _, err := new(jwt.Parser).ParseUnverified(string(body), jwt.MapClaims{})
+	if err != nil {
+		// Do NOT cache a partially-parsed token on error: a later cache hit reads
+		// its claims (the "exp" expiry) and would panic on a malformed token.
+		return nil, fmt.Errorf("failed to parse authentication token: %w", err)
+	}
 	c.SavedToken = token
 
-	return token, err
+	return token, nil
 }
 
 type LoginToken struct {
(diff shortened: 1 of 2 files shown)


chrlesur:
### Adversarial review trace

Pilot model: Claude. Independent reviewer model: Codex (gpt-5.5, read-only).

- Codex had flagged this exact latent bug during the #339 review (`JWT` returning `nil, nil` on a body-read error) and recommended filing it separately → #340.
- **DIFF review — GO**, no blocking findings. Confirms: read and parse errors now fail closed; no token is cached on error (removing both the nil-token deref panic and the malformed-token cache-poison panic); scope widening from the read-only case to the adjacent parse-error caching is justified (same function, same panic class, same invariant: never return/cache unusable auth material).
- Non-blocking minor (Codex): the cache-hit branch still reads `exp` with unguarded type assertions, so a syntactically valid token without a numeric `exp` would panic on the next call — a distinct case (valid token, not an error path), tracked separately in #342.

Gates green: `gofmt`, `go vet`, `staticcheck`, `go test ./internal/client -race`, coverage ratchet, `make build`. Tests are mutation-proven (both fix halves go RED when reverted).

PR #344, opened 2026-06-16
fix(#342): guard the exp claim in the JWT cache-hit (no panic on a malformed cached token)

chrlesur:
## Refs #342

The `JWT()` cache-hit branch read the cached token's expiry with **unguarded** type assertions:

```go
expireAt := c.SavedToken.Claims.(jwt.MapClaims)["exp"].(float64)
```

A syntactically valid token whose claims are `nil`, not a `jwt.MapClaims`, or whose `exp` is absent or non-numeric would **panic** there on the next call. (#340 stopped caching a token on a read/parse *error*; this closes the remaining case — a successfully-parsed token without a usable `exp`.)

## Change (`internal/client/api.go`)

Read the expiry with comma-ok assertions: an unusable cached token now **falls through to re-authentication** instead of panicking. A valid far-future token is still served from cache, unchanged.

## Tests (mutation-proven)

`TestJWTCacheHitGuardsExpClaim`: a far-future token is served from cache without re-auth (transport calls = 0); `nil` claims, a non-`MapClaims` type, missing `exp`, and a non-numeric `exp` each fall through to exactly one re-auth without panicking. Restoring the unguarded assertions turns the malformed-claims subtests RED.

Gates: `gofmt`, `go vet`, `staticcheck` (clean), `go test ./internal/client -race`, coverage ratchet (21.1%), `make build`.

With #340 + this change, `JWT()` can no longer panic on a malformed or expired-claim auth token — it fails closed or re-authenticates.

### DIFF
--- a/internal/client/api.go
+++ b/internal/client/api.go
@@ -287,11 +287,16 @@ func (c *Client) JWT(ctx context.Context) (*jwt.Token, error) {
 	defer c.lock.Unlock()
 
 	if c.SavedToken != nil {
-		expireAt := c.SavedToken.Claims.(jwt.MapClaims)["exp"].(float64)
-		tm := time.Unix(int64(expireAt), 0)
-
-		if time.Until(tm) > 5*time.Minute {
-			return c.SavedToken, nil
+		// Serve the cached token only while it stays valid for at least 5 more
+		// minutes. Read its expiry defensively: a token whose claims are nil, or
+		// whose "exp" is absent or non-numeric, is treated as unusable and falls
+		// through to re-authentication — never panicking on the type assertions.
+		if claims, ok := c.SavedToken.Claims.(jwt.MapClaims); ok {
+			if exp, ok := claims["exp"].(float64); ok {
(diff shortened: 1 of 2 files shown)


chrlesur:
### Adversarial review trace

Pilot model: Claude. Independent reviewer model: Codex (gpt-5.5, read-only). Codex recommended this exact fix (comma-ok guard) as a follow-up during the #340 review; #342 implements it.

**DIFF review — GO**, no blocking findings. Confirms the cache-hit is panic-proof for any cached token shape (nil claims, non-MapClaims, missing/non-numeric exp) while a valid far-future token is still served from cache; fall-through to re-auth is the right behaviour (the cache is an optimization, and re-auth is bounded per call). Codex's non-blocking suggestion (cover the non-MapClaims branch with a concrete type) was applied — a `jwt.RegisteredClaims` subtest was added.

Gates green: gofmt, go vet, staticcheck, `go test ./internal/client -race`, coverage ratchet (21.1%), make build. Tests mutation-proven (restoring the unguarded assertions turns the malformed-claims subtests RED).

Issue #350, opened 2026-06-16
OpenIaaS virtual_disk: connected drifts true on a powered-off VM (perpetual reconnect plan)


chrlesur:
## Symptom

After powering a VM off through Terraform, the next plan shows a perpetual in-place update on the attached data disk:

```
# cloudtemple_compute_iaas_opensource_virtual_disk.data will be updated in-place
  ~ connected = false -> true
Plan: 0 to add, 1 to change, 0 to destroy.
```

The plan never converges while the VM is off, and `terraform apply` would issue a `connect` (`PATCH .../connect`) against a halted VM.

## Root cause

`connected` is declared `Optional + Default: true` (an *intent*) in `internal/provider/resource_compute_iaas_opensource_virtual_disk.go`, but `Read` flattens it from the **runtime** VBD plug state (`vm.Connected`, via `FlattenOpenIaaSVirtualDisk`).

The OpenIaaS API distinguishes two notions:
- `connect`/`disconnect` (`PATCH .../connect|disconnect`) = **plug/unplug** (`currently_attached`, runtime) — this is what `connected` maps to;
- `attach`/`detach` (`PATCH .../attach|detach`) = the **VBD link** (logical attachment).

On a **halted** VM the platform reports every VBD as unplugged (`Connected=false`) — a runtime fact, not the user's intent. So `Read` writes `connected=false`, which conflicts with the `Default: true` intent → perpetual drift. The client's `OpenIaaSVirtualDiskConnection` exposes only a single `Connected bool`, so the provider cannot, from that field alone, tell "deliberately disconnected" from "VM powered off".

## Impact

**TF prod breaker**: any user who stops a VM via Terraform gets a non-empty plan on every subsequent run, and applying it plugs the disk back on a powered-off VM.

## Evidence

Captured on a live `tf vm` lifecycle run (create → stop → converge), 2026-06-16: VM `546cb3eb-e3fd-42f9-8460-40830e15b687`, data disk `d5b2e39b-f39e-48a9-bfaf-8afd2541d6b7`, convergence-after-stop plan = `~ connected = false -> true`. Create + convergence (VM on) was clean; destroy was clean (0 orphan).

## Proposed fix

[...]

PR #351, opened 2026-06-16
fix(#350): stop the data-disk connected drift on a powered-off VM

chrlesur:
Refs #350.

## Problem

After stopping a VM through Terraform, the next plan showed a perpetual in-place update on the attached data disk:

```
# cloudtemple_compute_iaas_opensource_virtual_disk.data will be updated in-place
  ~ connected = false -> true
Plan: 0 to add, 1 to change, 0 to destroy.
```

Applying it would issue a `connect` (`PATCH .../connect`) against a halted VM. **TF prod breaker**, captured on a live `tf vm` lifecycle run (create → stop → converge).

## Root cause

`connected` is `Optional + Default: true` (an *intent*), but `Read` flattened it from the **runtime** VBD plug state (`vm.Connected`). The OpenIaaS API distinguishes `connect`/`disconnect` (plug/unplug — `currently_attached`, runtime) from `attach`/`detach` (the VBD link). On a **halted/paused** VM the platform reports every VBD unplugged, so `Read` wrote `connected=false` while the intent stayed `true` → perpetual drift.

## Fix (Codex PLAN GO-WITH-CHANGES — Option D)

In `openIaasVirtualDiskRead`, after the pure flatten, override `connected` with a power-state-aware decision (`resolveOpenIaaSDiskConnected`):

- **VM running** → mirror the runtime (`vm.Connected`): a genuine disconnect still surfaces as drift;
- **VM halted/paused** → preserve the known intent (`d.Get("connected")`): a powered-off VBD unplug is **not** drift;
- **VM unreadable (403 → nil) / unknown power state** → **fail closed** with an actionable error;
- **disk readable but no longer attached to the configured VM** → attachment-drift error, never a plain `connected=false` (connect/disconnect cannot fix a detach).

The decision lives **outside** the pure flatten helper (it needs the VM power state and the prior intent). `os_disk.connected` on the VM resource is **untouched** (Computed-only, does not drift) — no scope creep. An id-only import (`virtual_machine_id` not yet known) keeps the flattened value until the configuration provides the VM (documented).

## Tests (mutation-proven)

- `resolveOpenIaaSDiskConnected` unit tests cover all 7 Codex cases; mutating the off-case to mirror the runtime makes the power-off tests RED.
- Wiring tests (HTTP mock) pin the end-to-end power-off-no-drift, running-mirrors-runtime, and fail-closed-on-unreadable-VM.

## Review & gates

- **PLAN Codex**: GO-WITH-CHANGES (Option D + fail-closed guards, applied).
- **DIFF Codex**: see review-trace comment.
- gofmt, vet, staticcheck, build, `go test -race`: green. Provider coverage 15.8% (up).

### DIFF
--- a/internal/provider/resource_compute_iaas_opensource_virtual_disk.go
+++ b/internal/provider/resource_compute_iaas_opensource_virtual_disk.go
@@ -2,6 +2,7 @@ package provider
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/cloud-temple/terraform-provider-cloudtemple/internal/client"
 	"github.com/cloud-temple/terraform-provider-cloudtemple/internal/provider/helpers"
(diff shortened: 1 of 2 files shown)


chrlesur:
## Codex review trace (gpt-5.5 high, read-only)

Verdicts recorded verbatim.

### PLAN review — **GO-WITH-CHANGES**
Option D (power-state-aware override) confirmed over Option F. Guards imposed and applied:
- decision OUTSIDE the pure flatten helper; override only `diskData["connected"]` in `openIaasVirtualDiskRead`;
- VM running → mirror runtime; halted/paused → preserve intent; VM nil(403)/error/unknown power state → fail closed; disk no longer attached to the configured VM → attachment-drift error (never a plain `connected=false`);
- `os_disk.connected` on the VM resource untouched (Computed-only, no scope creep).

### DIFF review — **GO** (no blocker)
> Blockers: none. No nil-deref or missed state-safety invariant in this diff.

All 8 attack questions confirmed: override placement, the 5-way decision, the VM-read fail-closed branches, `d.Get("connected")` intent + id-only skip, no nil-read regression, no Create/Update loop, and the 7 pure cases + 3 wiring tests.

Two non-blocking nits, applied: a `vmErr != nil` (non-403 HTTP error) wiring test, and an id-only (`virtual_machine_id == ""`) test proving no VM read is issued.

### Local gates
gofmt, vet, staticcheck, build, `go test -race`: green. Provider coverage 15.8% (up). The power-off-no-drift guard is mutation-proven (mirroring the runtime in the off-case makes the tests RED).

Issue #298, opened 2026-06-14
Schema golden gate: freeze and human-gate the provider state contract


chrlesur:
> **Plan reviewed by Codex (adversarial, 3 rounds): NO-GO → NO-GO → GO.**
> Round 1 added the missing state-affecting fields (TypeSet.Set, StateFunc, Default values, plan-constraints, explicit Elem kind, StateUpgrader version+type, CustomizeDiff, InternalValidate, CI-update ban). Round 2 added ValidateFunc/ValidateDiagFunc presence. Round 3: GO.

# Schema golden gate — freeze the externally-observable state contract

## Problem

The provider has no automated check that a code change does not silently alter
the **declared schema contract** that defines what an existing client's
Terraform state may contain. Line coverage proves nothing here. Freezing that
contract and failing CI on any change — forcing a human backward-compatibility
decision — is a high-leverage guard against silently breaking client state.

This **complements, and does not replace**, the runtime schema-vs-flatten
walker (#288) and the gating registries. It catches **declared schema drift
only**; runtime state corruption stays covered by the walker, resource tests,
plan-convergence work, and the integration suite (#8). See "Honest residual".

## Goal

A deterministic, machine-checked snapshot of the full provider schema, committed
to the repo, plus a CI test that fails on any divergence. Any contract change
must be seen and explicitly justified by a human before the golden is
regenerated.

## Scope (this increment)

1. **Schema serializer** (test-support code in `internal/provider`). From
   `New(version)()`, iterate `ResourcesMap` and `DataSourcesMap` (sorted keys).
   For each resource/datasource, serialize via its schema map, recursively for
   every attribute and nested `Elem` block. Per attribute, capture every field
   that can affect the stored state or whether an existing config still plans:
   - key (sorted),
   - `Type` string (TypeString/Bool/Int/Float/List/Set/Map),
   - `Required`, `Optional`, `Computed`, `ForceNew`, `Sensitive`,
   - `MinItems`, `MaxItems`, `ConfigMode`,
   - **`Default`**: serialize the literal value for primitives
     (string/bool/int/float); for non-primitive or `DefaultFunc`, record a
     `has_default_func` boolean. (Default *value* changes are breaking — record them.)
   - **state/normalization function PRESENCE** (booleans, never bodies):
     `has_state_func`, `has_diff_suppress_func`, `diff_suppress_on_refresh`,
[...]

Issue #257, opened 2026-06-12
Epic - Terraform Provider Cloud Temple : API convergence and state safety


chrlesur:
# Epic - Terraform Provider Cloud Temple : API convergence and state safety

## Why this epic

This epic tracks the convergence work between the Cloud Temple Terraform provider and the Cloud Temple API for OpenIaaS/VPC use cases.

The goal is not only to expose more API fields in Terraform. The primary constraint is Terraform state safety: schema, read, import, refresh and update changes must not unexpectedly corrupt existing client states, create permanent drift, or trigger destructive plans.

Internal support details, tenant-specific evidence and raw logs are tracked outside this public issue.

## Scope

In scope:

- OpenIaaS VM, disk and network adapter behavior.
- VPC datasource/resource coverage and relation to OpenIaaS resources.
- Terraform state-safety gates for provider changes.
- Existing provider issues and PRs related to OpenIaaS/VPC drift, retries, import/read parity and schema/API mismatch.
- Public, reproducible provider bugs or feature requests.

Out of scope:

- Tenant-specific incident details.
- Internal support or operations traces.
- Downstream consumer implementation details, except as validation feedback for the provider.

## Project tracking

Operational tracking is done in a private Cloud-Temple organization Project.

The Project contains prioritization, risk classification, owners, internal API/support items, and state-safety tracking. This public issue remains the public anchor for provider-facing work.

## Current public references

_Snapshot as of `rc/v1.8.0` @ `20239c0` (2026-06-16). The PR train referenced at epic creation is fully merged; see the progress tracker below for the per-workstream delivery log._

Merged onto `rc/v1.8.0` (referenced issues close at the RC release):

- The original train — #247, #249, #254, #256 (and #265, which superseded #252) — is merged.
- VPC provider surface — datasources (#301/#303/#304) and reversible write resources: `cloudtemple_vpc_static_ip` (#312), `cloudtemple_vpc_floating_ip_binding` (#314), VPC async-write resilience (#349).
- Client/auth resilience — read resilience + per-request timeout (#341), default timeout 240s→600s (#347), JWT fail-closed + `exp` guard (#343/#344).
- State-safety fixes — TF_LOG secret redaction (#321), OS-disk nil-read guards (#323/#326), empty SLA-assign hang (#307), empty cloud-init on create (#346), data-disk power-off `connected` drift (#351), schema golden gate (#299).
[...]

PR #352, opened 2026-06-16
test(#293): pin the VM StateUpgrader (V0->V1 extra_config migration)

chrlesur:
Refs #293. Refs #257.

## Problem

`migrateVirtualMachineStateV0toV1` is the **only** `StateUpgrader` in the provider, and it was **untested** — the highest state-safety gap in the S7 coverage roadmap (#293). A buggy StateUpgrader silently corrupts a client's Terraform state on an N-1 → N provider upgrade.

The V0 schema stored `extra_config` as a list of `{key, value}` objects; V1 stores it as a map. The upgrader converts the list to the map.

## Change (test-only)

A mutation-proven, table-driven unit test (`TestMigrateVirtualMachineStateV0toV1`) covering:

- V0 list → V1 map conversion;
- idempotence on already-migrated (map) state;
- absent / nil `extra_config` (no panic, no invented key);
- empty list → empty map;
- malformed items skipped without dropping the valid pairs or panicking (non-map item, missing key/value, non-string key/value);
- duplicate keys: last value wins;
- sibling state keys preserved across the migration.

No production code changes. The `internal/provider` coverage floor is raised 13.4 → 16.1 to lock in the gain (the upgrader was at 0%).

- Dropping the conversion (the `rawState["extra_config"] = extraConfigMap` assignment removed) → 5 cases RED.
- First-wins on duplicate keys → the duplicate case RED.

## Review & gates

- **DIFF Codex** (adversarial, read-only): **GO** — no blocking findings; the test is faithful to the function's real behavior (the `exists && != nil` guard, the `[]interface{}` type assertion, the per-item key/value type asserts, last-write-wins, and in-place sibling preservation are all verified). Optional non-blocking ideas (unexpected top-level type, nil item) are corrupted-state / out-of-SDK-pipeline cases. The `SchemaVersion` / `StateUpgrader` wiring itself is already frozen by the schema golden snapshot (#298/#299).
- Coverage ratchet: `internal/provider` 16.1% (floor raised to 16.1).

### DIFF
--- a/internal/provider/resource_compute_virtual_machine_unit_test.go
+++ b/internal/provider/resource_compute_virtual_machine_unit_test.go
@@ -1,6 +1,8 @@
 package provider
 
 import (
+	"context"
+	"reflect"
 	"testing"
 
 	"github.com/hashicorp/go-cty/cty"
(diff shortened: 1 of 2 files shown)


chrlesur:
## Codex DIFF review (adversarial, read-only) — VERDICT: GO

No blocking findings. The test faithfully reflects the real behavior of `migrateVirtualMachineStateV0toV1`.

Verified:

1. The expectations match the code: the `exists && extraConfigRaw != nil` guard then the `[]interface{}` type assertion produce exactly the tested behaviors.
2. Malformed case: the function silently skips non-`map[string]interface{}` items, missing keys/values, non-string keys and non-string values, while keeping `a -> 1` and `d -> 4`. Expected output correct.
3. Duplicates: sequential writes into the map, so last-write-wins; `{k: "2"}` is correct.
4. nil / absent / empty list: exact — nil and absent skip the conversion; an empty list becomes `map[string]interface{}{}`.
5. Already-migrated map: exact — not a `[]interface{}`, left unchanged.
6. Sibling keys: exact — the function mutates `rawState` in place and replaces only `extra_config`, so `id`/`name` are preserved.
7. Mutation proof credible: dropping the conversion reds the 5 list-based cases; first-wins reds the duplicate case. Not exhaustive, but it covers the obvious dangerous mutants.
8. No important missing case for the real V0 shape. Optional, non-blocking additions: unexpected top-level type, nil item, empty key, already-`map[string]string` — mostly corrupted-state / out-of-SDK-pipeline cases.
9. Floor `internal/provider 13.4 -> 16.1`: acceptable; the CI ratchet re-runs `go test -race -coverprofile` and would fail if 16.1 were too high.
10. Minor fragility: the test calls the function directly, not the full SDK pipeline — but the `SchemaVersion` / `StateUpgrader` wiring is already locked by the schema golden snapshot.

Conclusion: GO. Non-complacent test, faithful to current behavior, fit for the StateUpgrader risk.

PR #353, opened 2026-06-16
fix(#293): guard setIdFromActivityConcernedItems against an empty id + pin both id resolvers

chrlesur:
Refs #293. Refs #257.

## Problem

`setIdFromActivityState` and `setIdFromActivityConcernedItems` resolve a resource id from an async activity, and were **untested** (S7 coverage roadmap, #293). Setting the wrong — or empty — id makes Terraform track the wrong remote object, or makes the resource untrackable: a state-safety hazard.

A review also surfaced an **asymmetry**: `setIdFromActivityConcernedItems` would `SetId("")` when a matching `ConcernedItem` carried an empty `ID`, clobbering a previously-set id — whereas `setIdFromActivityState` already refuses an empty `Result`.

## Change

**Fix (1 line, production):** `setIdFromActivityConcernedItems` now refuses an empty id (`i > -1 && ConcernedItems[i].ID != ""`), symmetric with `setIdFromActivityState`. An empty id is not a trackable Terraform id, so not setting it is strictly safer than clobbering with `""`.

**Tests (mutation-proven, table-driven):**

- `setIdFromActivityState`: id set **only** on exactly one state with a non-empty `Result`; nil / empty / multi-state / empty-result are no-ops that never clobber a pre-existing id.
- `setIdFromActivityConcernedItems`: id from the item matching the expected type **regardless of position**, first-match wins on duplicates; no match / nil / empty / empty-id are no-ops that never clobber.

The `internal/provider` coverage floor is raised 16.1 → 16.4.

- Drop the exactly-one-state guard → the multi-state case REDs.
- Drop the not-found index guard → the no-match case panics (REDs).
- Drop the empty-id guard → the empty-id case REDs (clobber to `""`).

## Review & gates

- **DIFF Codex** (adversarial, read-only): GO on the technical change — the short-circuit guard is safe (no panic on no-match), the fix does not break a legitimate path, the tests faithfully reflect the corrected behavior, and the mutation proof covers the new guard. Known non-blocking residuals: a same-type duplicate where the first item has an empty id falls through to a no-op (the caller should fail closed on an empty `d.Id()`); several callers set the id before checking the `WaitForCompletion` error (pre-existing, not introduced here).
- Coverage ratchet: `internal/provider` 16.4% (floor raised to 16.4).

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -564,7 +564,9 @@ func setIdFromActivityConcernedItems(d *schema.ResourceData, activity *client.Ac
 	}
 
 	i := slices.IndexFunc(activity.ConcernedItems, func(concernedItem client.ActivityConcernedItem) bool { return concernedItem.Type == expectedType })
-	if i > -1 {
+	if i > -1 && activity.ConcernedItems[i].ID != "" {
+		// Symmetric with setIdFromActivityState: never adopt an empty id, which
+		// would clobber a previously-set id and make the resource untrackable.
 		d.SetId(activity.ConcernedItems[i].ID)
 	}
 }
--- a/internal/provider/provider_set_id_from_activity_unit_test.go
+++ b/internal/provider/provider_set_id_from_activity_unit_test.go
@@ -1,0 +1,193 @@
+package provider
+
+import (
+	"testing"
+
+	"github.com/cloud-temple/terraform-provider-cloudtemple/internal/client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// newEmptyResourceData returns a ResourceData backed by an empty schema. The
+// functions under test only touch SetId/Id, which are schema-independent.
+func newEmptyResourceData() *schema.ResourceData {
+	return (&schema.Resource{Schema: map[string]*schema.Schema{}}).TestResourceData()
+}
+
+// TestSetIdFromActivityState pins how a resource id is resolved from an async
+// activity's State (#293, S7). Setting the wrong id makes Terraform track the
+// wrong remote object — a state-safety hazard. The function must set the id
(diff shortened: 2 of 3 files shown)


chrlesur:
## Codex DIFF review (adversarial, read-only) — VERDICT: GO (technical)

Two rounds.

**Round 1 — GO-WITH-CHANGES.** Two findings, both addressed:
1. The "no-op never clobbers" invariant for `setIdFromActivityState` was only proven on the nil path → added the dangerous state-safety paths (prior id + empty result; prior id + multi-state).
2. Asymmetry: `setIdFromActivityConcernedItems` would `SetId("")` on a matching item with an empty `ID`, clobbering a prior id, whereas `setIdFromActivityState` refuses an empty `Result` → **fixed** with a symmetric empty-id guard, and pinned with an empty-id no-clobber case.

**Round 2 — GO (technical).** Verified:
1. The guard `i > -1 && ConcernedItems[i].ID != ""` is safe: Go `&&` short-circuits left-to-right, so `ConcernedItems[i]` is not evaluated when `i == -1` (no panic on no-match).
2. The fix breaks no legitimate path: an empty id is not a trackable Terraform id; not setting it is strictly safer than clobbering with `""`.
3. The tests faithfully reflect the corrected behavior; the added no-clobber cases and the empty-id case are correct.
4. Mutation proof covers the new guard (dropping the empty-id guard reds the empty-id case with a prior id).
5. The empty-id asymmetry is resolved.

Non-blocking residuals (not introduced by this change): a same-type duplicate whose first item has an empty id falls through to a no-op (the caller should fail closed on an empty `d.Id()`); several callers set the id before checking the `WaitForCompletion` error.

The only round-2 blocker raised was a process one — a forbidden tooling authorship trailer in the commit — which has been removed.

PR #354, opened 2026-06-16
test(#293): pin GetStringList (list attribute -> []string)

chrlesur:
Refs #293. Refs #257.

## Problem

`helpers.GetStringList` reads a Terraform list attribute into a `[]string` used to build the datasource filter payloads (`datacenters`, `networks`, `datastores`, `hosts`, `host_clusters`, `types`), and was untested. A wrong order, a dropped element, or a nil return would silently change what gets queried.

## Change (test-only)

A mutation-proven unit test covering:

- a configured list returns its strings in order;
- an empty list, an unset key, and a non-list key all fail closed to a **non-nil empty slice** (never nil, never a panic).

No production code changes. The `internal/provider/helpers` coverage floor is raised 68.1 → 70.0.

- Returning `nil` on the not-a-list path → the non-list case REDs.
- Dropping the copied value → the order case REDs.

## Review & gates

- **DIFF Codex** (adversarial, read-only): GO — assertions faithful to the real behavior, mutation proof credible. The only uncovered mutant (a non-string element panicking on `v.(string)`) is unreachable through the current SDK contract (all callers use `TypeList` + a `TypeString` Elem); noted, not blocking.
- Coverage ratchet: `internal/provider/helpers` 70.0% (floor raised to 70.0).

### DIFF
--- a/internal/provider/helpers/helper_get_string_list_test.go
+++ b/internal/provider/helpers/helper_get_string_list_test.go
@@ -1,0 +1,76 @@
+package helpers
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// TestGetStringList pins GetStringList (#293, S7). It reads a Terraform list
+// attribute into a []string for the datasource filter payloads. It must
+// preserve order and content, and must fail closed to a non-nil empty slice —
+// never nil, never a panic — when the key is unset or not a list. Callers feed
+// the result straight into API filter structs, so a wrong order, a dropped
+// element, or a nil return would silently change what gets queried.
+// Non-complacent: dropping the copied value reds the order case; returning nil
+// on the not-a-list path reds the non-list case.
+func TestGetStringList(t *testing.T) {
+	sm := map[string]*schema.Schema{
+		"tags": {
+			Type:     schema.TypeList,
+			Optional: true,
+			Elem:     &schema.Schema{Type: schema.TypeString},
+		},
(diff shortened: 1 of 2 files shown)


chrlesur:
## Codex DIFF review (adversarial, read-only) — VERDICT: GO

No blocking findings.

1. The assertions match the real behavior of `GetStringList`: a configured list → ordered copy; a non-list key → the `!ok` path → `[]string{}`; an unset list key → the SDK `TypeList` zero value (`[]interface{}{}`) → `make([]string, 0)`. No false positive.
2. The non-list key (`TypeString` `name`) correctly takes the `!ok` path (non-nil empty slice); the unset list key takes the OK path (empty list → `make([]string, 0)`, non-nil). Both non-nil assertions are correct.
3. Mutation proof credible for the contract: returning `nil` on `!ok` reds the non-list case; dropping the copied value reds the order case. The only uncovered mutant (a non-string element panicking on `v.(string)`) is unreachable through the current SDK contract — every caller uses a `TypeList` with a `TypeString` Elem (`datacenters`, `networks`, `datastores`, `hosts`, `host_clusters`, `types`). Reachable only by a future misuse of the helper, not by the current Terraform SDK contract.
4. `schema.TestResourceDataRaw` is a valid, stable harness here — it goes through the real SDK schema rather than a hand-built ResourceData.
5. The helpers floor bump 68.1 -> 70.0 is consistent with covering a previously-untested helper.
6. The commit carries no forbidden authorship trailer/footer.

Conclusion: GO.

Issue #356, opened 2026-06-17
OpenIaaS VM: provider-initiated reboot uses the stale planned host_id instead of the live host, risking an unintended relocation


chrlesur:
## Describe the bug

When `cloudtemple_compute_iaas_opensource_virtual_machine` performs a **provider-initiated reboot** during an in-place update (for example an `os_disk` / `os_network_adapter` change that requires a power-cycle, or a `power_state` change), the final power-on passes the planned/prior-state `host_id` (`d.Get("host_id")`) instead of the VM's **live current host**. Because `host_id` is `Optional+Computed`, when it is not set in the configuration this value can be stale, and the power-on can relocate the VM to a host the user never requested.

## Conditions to reproduce

- `host_id` is **not set** in the Terraform configuration (it stays `Computed`).
- The apply runs **without a refresh** (`terraform apply -refresh=false`, or a saved plan applied without refresh).
- The VM was migrated to another host **out-of-band** (e.g. XCP-ng HA / load-balancing) between the last refresh and this apply.
- An update triggers a provider-initiated reboot (disk resize, NIC change requiring a reboot, etc.).

## Result

The reboot powers the VM back on the **stale host recorded in state**, silently relocating a running VM across physical hosts as a side effect of an unrelated change. The apply reports success with no error.

## Code references (rc/v1.8.0)

- The device-update reboot power-on passes `HostId: d.Get("host_id")` (`internal/provider/resource_compute_iaas_opensource_virtual_machine.go`, ~line 1519) instead of the live `vm.Host.ID` already read earlier (~line 1345).
- The `power_state` block similarly passes `HostId: d.Get("host_id")` (~line 1031).
- The error-path restore (`powerOnBestEffort`, ~lines 1444 / 1470 / 1500) already uses the live `vm.Host.ID` — the success power-on is inconsistent with it.

## Proposed fix

When `host_id` is **not explicitly configured**, the provider-initiated power-on should target the VM's **live current host** (`vm.Host.ID`, already read), not the planned/stale `d.Get("host_id")` — aligning the success power-on with the existing `powerOnBestEffort` behavior. When `host_id` **is** explicitly configured, honor the configured host.

## Context

Discovered during the review of #355 (host_id update silent no-op). This is **distinct** from #355's reported symptom (an explicit `host_id` change that is silently ignored): it affects the case where `host_id` is unconfigured and state is stale. Pre-existing behavior in v1.7.x and rc/v1.8.0.

[...]

Issue #355, opened 2026-06-17
[BUG] `cloudtemple_compute_iaas_opensource_virtual_machine` — `host_id` update reports success but VM is not migrated (silent no-op)


mls-ops:
**Describe the bug**

When a `terraform apply` changes the `host_id` of a `cloudtemple_compute_iaas_opensource_virtual_machine` to migrate a VM to a different XCP-ng host **within the same cluster/pool** (intra-pool live migration), the provider reports `Modifications complete` with no error, but the VM is never actually moved. On the next `terraform plan`, the same `host_id` change reappears, creating an infinite apply loop with no observable effect on the infrastructure.

---

## Current Behavior

```
cloudtemple_compute_iaas_opensource_virtual_machine.workers["bestie-wrk-01"]: Modifying... [id=10cf8c47-f4e2-43eb-9997-1691b8f04b3d]
cloudtemple_compute_iaas_opensource_virtual_machine.workers["bestie-wrk-01"]: Modifications complete after 2s [id=10cf8c47-f4e2-43eb-9997-1691b8f04b3d]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

The apply exits 0 with no error. However:
- The VM (`bestie-wrk-01`, UUID `10cf8c47-f4e2-43eb-9997-1691b8f04b3d`) remains on its original host (`xcp-hk64-ucs17-az05`, host UUID `723f5189-6321-49ea-bb2a-e26a9a82fbaa`) as confirmed in XenOrchestra.
- A subsequent `terraform plan` shows the same `host_id` change again:

```
~ host_id = "723f5189-6321-49ea-bb2a-e26a9a82fbaa" -> "f9906934-7d68-448d-8c72-2d57277269a1"
```

This creates an **infinite apply loop**: every apply succeeds (exit 0) but no migration takes place.

---

**Expected behavior**

The provider should either:
1. Successfully trigger the XCP-ng/XenOrchestra live migration and wait for its completion before reporting success, OR
2. Return an explicit error if the migration cannot be performed (e.g. VM is powered on and live migration is not supported in the given configuration).

Under no circumstance should the provider silently report success while leaving the VM on its original host.

---

**Steps to reproduce**

1. Have a `cloudtemple_compute_iaas_opensource_virtual_machine` running on host A within **XCP-ng cluster X** (intra-pool migration).
[...]

PR #357, opened 2026-06-17
fix(#355): migrate OpenIaaS VM host_id changes via the relocate API

chrlesur:
## What

Fixes the silent no-op when changing `host_id` to migrate a running `cloudtemple_compute_iaas_opensource_virtual_machine` to another host in the same pool: the apply reported `Modifications complete` while the VM never moved, and every subsequent plan replayed the same `host_id` change (an endless apply loop).

## How

- New client method `OpenIaaSVirtualMachineClient.Relocate` (`POST /compute/v1/open_iaas/virtual_machines/{id}/relocate`, body `{hostId}`).
- Host placement is decided from raw-config intent only — a value merely computed from the live API is never treated as a migration request.
- A `host_id` change that would leave the VM powered off is rejected up front, before any mutation, in both `Create` and `Update`.
- On a running VM, the change triggers the relocate (idempotent, and only on positive live evidence the VM is actually running), waits for the migration activity, then asserts the VM converged onto the requested host (live power `on`, live host == requested) — failing closed with an explicit diagnostic otherwise, instead of ever reporting a false success.

## Tests

- Pure decision + raw-config table tests; a stateful-fake orchestration test (mutation-proven: removing the relocate call, the convergence check or the live-power guard goes red); an `httptest` client test for `Relocate`; an acceptance test for the powered-off preflight.
- `gofmt` / `go vet` / `staticcheck` clean; `go test -race -cover` green (client 21.5%, provider 17.0%, helpers 70.0%, all above the CI floors).

## Scope

The pre-existing unconfigured-host stale-power-path behaviour (the legacy power paths sending `d.Get("host_id")` under `-refresh=false`) is tracked separately in #356 and intentionally left untouched here.

Refs #355


chrlesur:
## Independent reviewer pass on the committed diff (Codex, gpt-5.5, read-only)

Recorded verbatim.

```
VERDICT: GO

BLOCKING: None.

MINOR:
- The acceptance test proves the Create fail-fast path; Update fail-fast is verified by code inspection here, not by a dedicated call-site test. Acceptable for this PR because `placementInputs` and preflight are visibly at the top of Update before any client mutation.

NITS: None.

Attack questions:
1. Create and Update preflight are before any client mutation. Update captures `placementInputs` once and reuses them in `applyOpenIaaSHostPlacement`.
2. No in-scope path silently succeeds with live host != requested host; relocate, off->on, and create all end in live convergence checks.
3. Relocate is only issued after live `power == "on"` evidence.
4. Power-block semantics are preserved: same off-skip, wait, driver wait, and error propagation, just wrapped as `error`.
5. No relocate `SetId`; `host_id` remains Optional+Computed, not ForceNew; schema golden untouched.
6. #356 is correctly left untouched: the stale unconfigured-host power paths remain as accepted out-of-scope debt.
7. Tests are not complacent: fake is stateful, convergence/relocate/live-power guards have RED-without-fix style coverage per evidence.
8. No regression seen for replication, properties reboot, mount_iso, boot_order, or device-update paths; ordering is unchanged except the old power block is now wrapped and followed by convergence.
```

_Plan review: GO after a multi-round adversarial pass. Pre-commit review: GO. Full local gate green (gofmt, go vet, staticcheck, `go test -race -cover`). Merge held pending human approval._

Issue #359, opened 2026-06-18
platform/api: OpenIaaS — auto-created "xoa" VPC static IP is not released when its network adapter/VM is deleted (orphan accumulation)


chrlesur:
Surfaced on the OpenIaaS recette tenant during repeated VM lifecycle runs (`ct-validate compute_lifecycle` soak). This is the platform-side counterpart of the provider doctrine in #311 (the provider deliberately refuses to manage `xoa` static IPs), and is related to the VPC contract issues in #317.

## Symptom

After repeated create/destroy of OpenIaaS VMs whose network adapter is attached to a VPC-backed network, the private networks accumulate **orphaned static IPs**: each one references a `virtualMachine` / `networkAdapter` that no longer exists, and none of them is ever cleaned up. They survive every VM teardown and pile up run after run.

## Mechanism

1. Attaching a network adapter to a VPC-backed network makes the platform **auto-create a `source: "xoa"` static IP** for that adapter's MAC (it then appears in `GET /vpc/v1/private_networks/{id}/static_ips`). This is the platform's own mapping, not a client-created (`custom`) one.
2. When the adapter (and/or the VM) is deleted, the platform **does not release that xoa static IP**. It remains on the private network, still pointing at the now-deleted `virtualMachine` / `networkAdapter`.
3. The orphan **cannot be removed via the API**: `DELETE /vpc/v1/static_ips/{id}` refuses it with *"not a custom static IP"* (only `source: "custom"` IPs are deletable). So nothing — neither the provider nor any API client — can clean it up; only platform-side GC or the web console can.

## Live evidence (OpenIaaS recette tenant, VPC `FW_AGENTIC`, 2026-06-18)

The machine manager reports **0 VMs**, yet **7 static IPs** remain, all `source: xoa`, each referencing an **absent** VM:

| private network | IP | MAC | source | referenced VM | floating IP |
|---|---|---|---|---|---|
| DMZ | 192.168.191.10 | 32:c1:e2:ae:f9:75 | xoa | d1ae8c75-… (absent) | — |
| DMZ | 192.168.191.11 | 56:94:06:2b:a0:6e | xoa | 9b5ce279-… (absent) | — |
| LAN | 192.168.190.10 | b6:fa:75:c9:45:62 | xoa | eefade41-… (absent) | — |
| LAN | 192.168.190.11 | 26:c1:29:ab:6d:da | xoa | 84c483e4-… (absent) | — |
[...]

PR #363, opened 2026-06-22
docs(ct-test): add TF VM relaunch runbook (+ #359 xoa side-effect caveat)

chrlesur:
## What

Adds `examples/ct-test/vm/RUNBOOK.md` — a focused, operational "how to relaunch
`ct-test.sh tf vm`" runbook (quick command + `CT_ENV_OPENIAAS` override, prerequisites,
lifecycle, tunables, troubleshooting table). It **complements** the existing `README.md`
(which describes what the scenario exercises) rather than duplicating it, and is
cross-linked from it.

## Why now

Documents a **known side effect** surfaced while diagnosing #359: the scenario attaches
an adapter to the VPC-backed LAN, so the platform auto-creates an `xoa` static IP that
`destroy` does not release — **each run leaves one orphaned `xoa` static IP on LAN**.
This is not a test failure (the smoke check does not look at it), but it accumulates, so
re-runners need to know to audit/purge separately.

Docs only — no code change.

Refs #316

### DIFF
--- a/examples/ct-test/vm/README.md
+++ b/examples/ct-test/vm/README.md
@@ -5,6 +5,9 @@ API): deploy an Ubuntu VM **from the marketplace**, attach a data disk, then dri
 power (start → stop → start), with a **convergence check** (empty `terraform plan`)
 after every step and a clean `destroy` at the end.
 
+> Relaunching it? See [RUNBOOK.md](RUNBOOK.md) for the quick checklist, prerequisites,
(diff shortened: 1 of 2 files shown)


pbesret:
Closing without merging: this runbook was written around the #359 xoa static-IP side-effect that was believed to be a provider bug, but it turned out to be a not-yet-ready API rather than a defect in the provider. The runbook is no longer relevant, so this PR is being closed.

PR #362, opened 2026-06-22
chore(#238): remove unpublished VPC provider surface

pbesret:
Refs #238

## Why
The `cloudtemple_vpc_*` provider surface targets `/vpc/v1`, a **deprecated** API contract that the platform is replacing with breaking changes **under the same URL** (no `/v2` coexistence). This surface was **never published** in a release. Shipping it in v1.8.0 would expose a client surface that would break server-side shortly after release, with no recourse for users. v1.8.0 must remain a **security + bug-fix** release.

## What
- **Remove the user-facing VPC surface** (Layer A): the VPC datasources, the `cloudtemple_vpc_static_ip` and `cloudtemple_vpc_floating_ip_binding` resources, their flatten helpers and tests, and the provider registrations. The corresponding CHANGELOG entries are removed (the v1.8.0 section no longer advertises any VPC feature).
- **Quarantine the internal VPC client SDK** (Layer B, `internal/client/vpc*.go`): it still compiles and is exercised only by the opt-in `ct-validate vpc` cycle; it is not wired into any shipped provider resource/datasource.
- **Kept on purpose:** the OpenIaaS network-adapter attributes `vpc_id` / `vpc_name` / `private_network_*` / `static_ip_address`. They are sourced from the **compute** API (`FlattenOpenIaaSNetworkAdapter`), not from `/vpc/v1`, so they are unaffected by the deprecated contract.

## Scope / issues
This is a feature PR onto the RC branch, so it uses `Refs #238` (no closing keyword, per the RC flow). The VPC issues (#238, #308, #309, #319, #348) stay **open** and are **deferred to v1.9.0**, where the surface will be rebuilt against the new `/vpc/v1` contract. They must NOT be closed by the v1.8.0 RC → main PR.

## Evidence
- CI `Tests` green on the tip commit `2afbc94` (build + unit/state-safety suites + `go vet` + staticcheck + race detector + datasource schema-vs-flatten validation + schema golden gate + coverage ratchet).
- No user-facing VPC reference remains: no `docs/`, `examples/` or provider VPC files; no `cloudtemple_vpc` entry in the CHANGELOG. The only residual VPC code is the quarantined internal client and the internal `cmd/ct-validate/cycle_vpc.go` harness (not part of the shipped provider surface).

## Review & merge
Feature → RC. Independent adversarial review (Codex) is being run; the verdict will be posted on this PR. Merge is gated on a favorable verdict **and** an explicit human GO (shared/integration branch).

Issue #320, opened 2026-06-15
panic: runtime error: invalid memory address or nil pointer dereference in FlattenOpenIaaSOSDiskData


mls-ops:
**Describe the bug**
During a `terraform plan` execution in the terraform workspace, the `cloud-temple/cloudtemple` provider (version `v1.7.1`) crashed with a `nil pointer dereference` when refreshing state for virtual machines. Specifically, the crash traceback points to `FlattenOpenIaaSOSDiskData` when processing a VM's OS disk.

## Current Behavior
The provider crashes during `terraform plan` and prints the following panic stack trace:

```text
Stack trace from the terraform-provider-cloudtemple_v1.7.1 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x78 pc=0x955472]

goroutine 1041 [running]:
github.com/cloud-temple/terraform-provider-cloudtemple/internal/provider/helpers.FlattenOpenIaaSOSDiskData(0x0, {0xc00086bf80, 0x24})
	github.com/cloud-temple/terraform-provider-cloudtemple/internal/provider/helpers/helper_compute_iaas_opensource_virtual_machine.go:86 +0x32
github.com/cloud-temple/terraform-provider-cloudtemple/internal/provider.openIaasVirtualMachineRead({0xf60040, 0xc0004bcfc0}, 0xc000533d00, {0xdbd700?, 0xc0001bf710})
	github.com/cloud-temple/terraform-provider-cloudtemple/internal/provider/resource_compute_iaas_opensource_virtual_machine.go:631 +0x607
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).read(0xc000425c00, {0xf5ff98, 0xc00062dad0}, 0xc000533d00, {0xdbd700, 0xc0001bf710})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.35.0/helper/schema/resource.go:823 +0x119
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).RefreshWithoutUpgrade(0xc000425c00, {0xf5ff98, 0xc00062dad0}, 0xc000649c70, {0xdbd700, 0xc0001bf710})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.35.0/helper/schema/resource.go:1117 +0x525
[...]

Issue #319, opened 2026-06-15
feat: bounded transient-retry for vpc_floating_ip_binding bind/unbind (resilience)


chrlesur:
Refs #309, #315. Follow-up agreed when #309 (`cloudtemple_vpc_floating_ip_binding`) was merged: make bind/unbind resilient to transient platform activity failures.

## Context

On 2026-06-15 the VPC write path intermittently failed with `[ComputeManager] None of the workers were able to respond` (transient), so a single `terraform apply` failed and then succeeded on retry. The provider already recognises this reason (`client.IsTransientActivityFailure`) and retries it for VIF updates (#251, `runVIFUpdateWithRetry`), but `vpc_floating_ip_binding` bind/unbind do not yet.

## Ask

Add a bounded retry on transient activity failures to `Bind`/`Unbind` (the #251 pattern): re-read before each attempt (already-converged → success), retry only `IsTransientActivityFailure`, bounded attempts, no hardcoded sleeps (injectable backoff), mutation-proven tests. Preserve the never-drop / never-orphan invariants.

pbesret:
Closing — the concrete trigger for this client-side bounded retry was the transient VPC activity failure `[ComputeManager] None of the workers were able to respond` (#315), which has been resolved API-side. With the platform no longer emitting those transient failures on the VPC write path, a dedicated bounded-retry on `vpc_floating_ip_binding` bind/unbind is unnecessary. If transient binding failures recur in practice, reopen and apply the #251 pattern.

Issue #309, opened 2026-06-14
feat(vpc): cloudtemple_vpc_floating_ip_binding — public IP bind/unbind (VPC API)


chrlesur:
Related to #238 (VPC provider surface, public IP).

A **binding** resource for the public IP, reversible by construction (the floating IP itself is pre-provisioned out-of-band — no create/destroy of the FIP here):
- Create = `POST /vpc/v1/floating_ips/{id}/bind/static_ips/{staticIpId}` (bind, async).
- Delete = `DELETE /vpc/v1/floating_ips/{id}/unbind/static_ips/{staticIpId}` (unbind, async).
- (Optional) manage the FIP `description` via `PATCH /vpc/v1/floating_ips/{id}`.

This is the day-to-day 'expose a VM publicly' action: VM adapter (MAC) → static IP (private) → floating IP (public). Depends on `cloudtemple_vpc_static_ip` existing.

NB: the FIP **provisioning** (Order API, by count, no deprovision) is a separate, API-blocked concern — see the gateway/FIP-provisioning issue. Non-complacent tests, golden additive, target rc/v1.8.0.

pbesret:
Delivered by #373 (C5, merged to rc/v1.9.0): `cloudtemple_vpc_floating_ip_binding`. Live E2E on a real tenant passed all scenarios (create+idempotency, anti-clobber refusal, import round-trip, destroy/unbind, 0-orphan sweep); CI green. Closing manually (base rc/v1.9.0 does not auto-close).

PR #372, opened 2026-06-26
Release v1.8.0

### LINKED ISSUE #241: Error unmarshalling data.cloudtemple_compute_virtual_machine

### LINKED ISSUE #243: OpenIaaS : datasource iaas_opensource_pools inutilisable en listing (Invalid address to set: pools.0.label)
## Problème

`data "cloudtemple_compute_iaas_opensource_pools" "all" {}` échoue systématiquement :

```text
Error: Invalid address to set: []string{"pools", "0", "label"}
```

Provider **1.7.1**, Terraform **1.15.5**, tenant OpenIaaS (AZ05).

## Reproduction minimale

```hcl
terraform {
  required_providers {
    cloudtemple = { source = "Cloud-Temple/cloudtemple", version = "~> 1.7" }
  }
}
provider "cloudtemple" {}
data "cloudtemple_compute_iaas_opensource_pools" "all" {}
```

`terraform apply` (read-only) → erreur ci-dessus. Les autres datasources de listing OpenIaaS (`availability_zones`, `networks`, `templates`, `storage_repositories`, `virtual_machines`) fonctionnent sur le même tenant.

## Analyse probable

Le schéma de la datasource ne déclare pas l'attribut `label` que l'API renvoie pour chaque pool (mismatch schéma ↔ réponse API lors du `d.Set`).

## Contournement utilisé

Cibler un pool par `id` via la datasource singulière.

Contexte projet : Cloud-Temple/agentic-platform#13.

### LINKED ISSUE #245: OpenIaaS : HTTP 500 transitoires sur les lectures API (replication policy, virtual_disks) — retry/backoff côté provider
## Contexte

Provider `cloudtemple` 1.7.1, tenant OpenIaaS. Lors de cycles répétés de déploiement/destruction pilotés par Terraform, nous observons des **HTTP 500 transitoires récurrents** sur des opérations de lecture, qui font échouer le run alors qu'un simple retry suffit.

## Occurrences observées (sur quelques jours de tests)

- Lecture de la *replication policy* pendant un `destroy` de VM (déjà signalé comme contrainte de runbook) — systématique ou quasi.
- Datasource `cloudtemple_compute_iaas_opensource_virtual_disks` (listing) — plusieurs occurrences, dont une en pleine photographie d'inventaire post-déploiement.
- Plus sporadiquement, d'autres datasources d'inventaire.

Dans tous les cas, un retry immédiat (1 à 2 tentatives, quelques secondes) réussit : ce sont des 500 transitoires côté API, pas des erreurs persistantes.

## Comportement attendu

Le provider devrait intégrer un **retry avec backoff** sur les réponses 5xx des opérations de lecture idempotentes (GET datasources, lectures intermédiaires des CRUD), comme le font la plupart des providers (AWS, etc.). Aujourd'hui chaque consommateur doit wrapper Terraform dans des scripts de retry, ce qui masque les vraies erreurs.

## Questions

1. Un mécanisme de retry est-il prévu côté SDK/provider ?
2. Ces 500 transitoires sont-ils connus côté API (nous les signalons en parallèle au support) ?

Liée : #243 (datasource pools).

### LINKED ISSUE #246: OpenIaaS : update VIF redondant après un deploy marketplace (create → update inconditionnel) transforme une panne annexe en échec total
## Contexte

Provider 1.7.1 (et `main` au commit 03b5e64), ressource `cloudtemple_compute_iaas_opensource_virtual_machine` créée depuis un item marketplace.

## Problème

À la création depuis la marketplace, le réseau cible de chaque adaptateur est **déjà transmis dans la requête de deploy** via `NetworkData` (mapping réseau source → réseau destination, `resource_compute_iaas_opensource_virtual_machine.go` l.514-521).

Pourtant, `openIaasVirtualMachineCreate` se termine par `return openIaasVirtualMachineUpdate(...)` (l.587), et au sein d'un create Terraform, `d.HasChange("os_network_adapter")` est **toujours vrai** (ressource neuve). Résultat : le provider enchaîne systématiquement un appel `NetworkAdapter().Update` (« Update network adapter VIF #0 ») pour chaque adaptateur (l.870-877 → `handleUpdateOSDevices` → `osNetworkAdapterUpdate` l.1063), **avant le power-on**, alors que le deploy vient d'appliquer exactement la même configuration.

## Impact

- Un appel API supplémentaire par VIF et par VM créée, qui ne change rien en nominal.
- Si l'opération « update VIF » échoue côté plateforme (notre cas : activité en échec `VPC configuration incomplete: None of the workers were able to respond`, 7 VMs sur 7), **toute la création échoue et la VM reste taintée**, alors que le deploy lui-même (avec le bon réseau transmis dans `NetworkData`) avait réussi. L'opération redondante transforme un bug d'une opération annexe en échec total de provisioning.

## Comportement attendu

Après le deploy, l'état réel des adaptateurs est déjà relu et posé dans le state (l.555-567). Suggestion : ne déclencher `osNetworkAdapterUpdate` à la création que si la configuration demandée diffère réellement de l'état remonté par l'API (network_id différent, mac_address explicitement fixée, tx_checksumming explicite), plutôt que sur `HasChange` qui est toujours vrai pour une ressource neuve.

Liées : #244 (multi-NIC marketplace), #245 (retries 5xx).

### LINKED ISSUE #251: OpenIaaS : retry borné des opérations VIF dont l'activité échoue pour une raison transitoire (« None of the workers were able to respond »)
## Problème

Les opérations sur les network adapters (create/update) passent par les workers VPC. Quand ceux-ci sont temporairement muets, l'activité échoue avec :

> VPC configuration incomplete: None of the workers were able to respond. We invite you to contact support.

Constaté le 11/06 à 16:27 (7/7 VMs, incident initial) puis à nouveau à 22:26 (1 création de VIF sur un run par ailleurs sain) : c'est **transitoire et récurrent**. Aujourd'hui le provider abandonne au premier échec → ressource taintée, apply en échec, alors qu'un retry quelques secondes plus tard réussit.

## Attendu

- Reconnaître les raisons d'échec d'activité transitoires (marqueur « None of the workers were able to respond ») et **retenter l'opération de manière bornée** (ex. 3 tentatives, backoff).
- Sur un retry de **création** de VIF : vérifier d'abord par listing si l'adaptateur a finalement été créé par la tentative précédente (adoption au lieu de doublon).
- Les raisons permanentes (conflit MAC, espace insuffisant, etc.) restent immédiatement fatales.

## Contexte

- Complémentaire de #245 (500 transitoires sur lectures) : ici c'est l'**état final de l'activité** qui est en échec transitoire, pas le transport.
- Volet plateforme (cause racine des workers muets) : déjà au support + #250 (idempotence Static IP).

Liées : #245, #246, #248, #250.

### LINKED ISSUE #253: OpenIaaS : boot_firmware non Computed → drift permanent (uefi → null) sur toute VM marketplace, le plan ne converge jamais
## Problème

Le champ `boot_firmware` de `cloudtemple_compute_iaas_opensource_virtual_machine` est `Optional` mais **pas `Computed`**. Les images marketplace posent `uefi` ; quand l'utilisateur ne configure pas le champ, chaque `terraform plan` post-apply propose :

```
- boot_firmware = "uefi" -> null
```

…sur **toutes** les VMs (constaté 11/11 sur notre stack, run h4ow1o) : le plan d'idempotence n'est jamais vide, ce qui casse tout pipeline qui vérifie la convergence (`plan -detailed-exitcode`).

## Fix

Ajouter `Computed: true` au schéma de `boot_firmware` (même pattern que `secure_boot`, `host_id`, etc.).

Constat : run de validation du 11/06 (cf. #246/#247 pour le contexte).

### LINKED ISSUE #255: OpenIaaS : le cloud-init config drive (XO CloudConfigDrive) peut être capturé comme os_disk à la création → drift de suppression permanent
## Problème

À la création d'une VM, le provider liste les disques de la VM et pose le résultat en `os_disk` dans le state. Selon le timing, le **config drive cloud-init** (`XO CloudConfigDrive`, 10 MiB, attaché en lecture seule par XO au deploy) est déjà présent et se retrouve capturé comme un `os_disk` géré.

Conséquence : tout plan suivant propose de **retirer** ce bloc (`- os_disk { name = \"XO CloudConfigDrive\" ... }`) — drift permanent, plan jamais convergent. Constaté sur une VM recréée en retry (run `tqg1ub`, git-01) : la course est gagnée ou perdue selon les délais plateforme.

## Fix proposé

Exclure du flatten `os_disk` les disques plateforme : VBD **ReadOnly** pour la VM concernée (critère structurel) et/ou nom `XO CloudConfigDrive`. Ce disque est géré par la plateforme (cloud-init), jamais par l'utilisateur.

Constaté pendant la validation de #247 (runs des cycles #21). Liées : #246, #253 (même famille : champs/objets plateforme non modélisés qui cassent la convergence).

### LINKED ISSUE #258: Governance: add provider workflow rules under RULES/WORKFLOW_GIT.md
Part of #257.

## Goal

Add provider-local workflow rules under `RULES/WORKFLOW_GIT.md`.

The provider repository should carry its own operating rules so contributors do not have to infer them from another workspace.

## Expected content

The initial `RULES/WORKFLOW_GIT.md` should cover:

- all changes to `main` go through GitHub pull requests;
- feature work starts from a dedicated branch;
- PRs that resolve an issue include `Closes #N` in the PR body;
- design discussion stays in the issue, code review discussion stays in the PR;
- GitHub Project `Status` represents workflow state only, not lot, risk, priority, or audit category;
- provider changes must be reviewed for Terraform state safety before merge;
- model-assisted work separates pilot and reviewer roles when available;
- urgent provider fixes are not blocked by this governance task.

## Acceptance criteria

- A `RULES/` directory exists in this repository.
- `RULES/WORKFLOW_GIT.md` documents the workflow above in provider-specific terms.
- The implementing PR references this issue with `Closes #<issue-number>`.
- The PR records which model acted as pilot and which model acted as reviewer, if model assistance is used.
- The PR receives normal GitHub review before merge.

## Notes

This is a governance/stabilization task. It does not change provider behavior.

### LINKED ISSUE #260: docs: refresh root README (outdated requirements, missing registry links, typos)
The root README.md is out of date:

- Requirements list Go >= 1.18 while go.mod requires Go 1.22 (toolchain go1.23).
- No link to the Terraform Registry page (Cloud-Temple/cloudtemple) or to the provider documentation.
- The usage example lacks the `required_providers` block and uses realistic-looking credential UUIDs instead of obvious placeholders.
- The provider arguments list is missing `api_suffix` and is not aligned with docs/index.md.
- Several typos in the acceptance-tests section ("whant", "testes", "lunch").
- No mention of the MPL-2.0 license or link to the CHANGELOG.

Refresh the README accordingly. Documentation-only change, no code impact.

### LINKED ISSUE #262: Security: remediate open Dependabot alerts on Go dependencies
9 open Dependabot alerts on the default branch (1 critical, 2 high, 4 moderate, 2 low), all runtime Go dependencies:

- **CRITICAL** — `google.golang.org/grpc` < 1.79.3 (CVE-2026-33186) → bump to v1.79.3
- **HIGH** — `github.com/dgrijalva/jwt-go` <= 3.2.0 (CVE-2020-26160): no patched release (archived project) → migrate to the maintained fork `github.com/golang-jwt/jwt/v4` (drop-in replacement; single usage in `internal/client/api.go`)
- **HIGH + 2 MEDIUM** — `golang.org/x/crypto` < 0.45.0 (CVE-2025-22869, CVE-2025-47914, CVE-2025-58181) → bump to v0.45.0
- **2 MEDIUM** — `golang.org/x/net` < 0.38.0 (CVE-2025-22870, CVE-2025-22872) → bump to v0.38.0
- **2 LOW** — `github.com/cloudflare/circl` < 1.6.3 (CVE-2025-8556, CVE-2026-1229) → bump to v1.6.3

Remediation targets the `rc/v1.8.0` integration branch (RC train).

Housekeeping note: the historical Dependabot PRs #129, #121, #111, #110, #109 and #108 (2023-2024) are outdated — e.g. #129 bumps terraform-plugin-sdk/v2 to 2.32.0 while the repository is already on 2.35.0 — and should be reviewed/closed by a maintainer.

### LINKED ISSUE #264: CI: establish a platform-independent unit-test and static-analysis safety net

### LINKED ISSUE #267: OpenIaaS: gate the VM-properties and VIF PATCH payloads on real live divergence

### LINKED ISSUE #269: Datasources: systematic schema-vs-flatten validation (the #243 class)
Lot C of the non-complacent test plan (#264).

## Problem

The flatten helpers and the datasource schemas can drift independently: when a helper emits a key the schema does not declare, `d.Set` fails at runtime with `Invalid address to set` and the datasource becomes unusable. Proven instance: #243 — `FlattenOpenIaaSPool` emits `label`, `master` and `memory`, none of which are declared in the `pools` Elem schema of `cloudtemple_compute_iaas_opensource_pools`.

Nothing protects the other ~90 datasources from the same class.

## Changes

1. A reusable test harness: every registered datasource gets its flatten output (built from fully-populated client fixtures) validated against its schema via `d.Set` — any undeclared path, type mismatch or list-vs-object drift fails the unit job with the exact address.
2. Seed the registry with the OpenIaaS datasources (highest exposure); adding an entry is a one-liner so the coverage grows with each future datasource.
3. Fix the mismatches the walker surfaces — at minimum the #243 pool schema (additive: declare the missing computed attributes).
4. Pin the `UpdateMapItems`/`GetOk` explicit-false behavior in the helpers tests (root cause class of the swallowed booleans, documented mitigation = raw-config gating).

Refs #243, refs #264.

### LINKED ISSUE #271: State-safety: VMware boot_options gating, secrets preserved on refresh, read errors are not deletions

### LINKED ISSUE #273: OpenIaaS: clean the legacy config-drive pollution from os_disk on refresh, and stop crashing on deleted devices

### LINKED ISSUE #281: State-safety: non-OpenIaaS resources must not drop from state on an unconfirmed read
## Problem

Several resource `Read` paths remove the resource from the Terraform state (`d.SetId("")`) as soon as the API client returns `nil` for a per-id read. But the client maps some HTTP codes to `nil` that do **not** prove the resource was deleted:

- most resources map **HTTP 403** to "not found" (`requireNotFoundOrOK(resp, 403)`), so a permission/scope condition is indistinguishable from a real deletion;
- `cloudtemple_iam_personal_access_token` maps **HTTP 500** to "not found", so a transient server error is indistinguishable from a real deletion.

As a result, a transient error, a permission hiccup or a credentials/scope change can **silently shrink a client's Terraform state**. On the next apply the resource is recreated, which for a personal access token also rotates the secret and can leave the previous credential live platform-side.

The OpenIaaS resources were already hardened against this class (strict, complete-listing deletion evidence). This issue extends the same doctrine to the remaining resources.

## Doctrine

A resource must never be removed from the state without **positive, unambiguous proof** of deletion. When such proof cannot be established, the provider fails closed: it returns a diagnostic and leaves the resource untouched in the state, rather than dropping it.

## Scope (per-resource qualification, delivered as separate PRs)

- [x] **Lot A** — `cloudtemple_iam_personal_access_token` (500→nil). The token listing scope is not provably equal to the resource's ownership scope, so absence is not deletion evidence: the resource never auto-removes itself; an unconfirmed read fails closed. _(PR #282)_
- [x] **Lot B** — `cloudtemple_compute_virtual_machine`, `cloudtemple_compute_virtual_disk`, `cloudtemple_compute_network_adapter`, `cloudtemple_compute_virtual_controller` (403→nil). Strict listing variants (`ListStrict`, HTTP 200 only) were added; the resources never auto-remove themselves on an unconfirmed read — a still-listed resource fails closed (no stale-state success), an absent or unreadable one also fails closed. _(PR #283)_
- [x] **Lot C** — `cloudtemple_object_storage_bucket`, `cloudtemple_object_storage_storage_account` (403→nil). The listings are unscoped and not provably complete, so absence is never treated as a deletion; every inconclusive read fails closed. _(PR #285)_

Each lot ships with non-complacent unit tests in the CI-covered packages.

A related, pre-existing hardening (per-id reads accepting a partial HTTP 206) is tracked separately in #286.

### LINKED ISSUE #288: Extend the datasource schema-vs-flatten walker to every datasource

### LINKED ISSUE #293: Non-complacent unit-test coverage for the pure / state-facing layer

### LINKED ISSUE #298: Schema golden gate: freeze and human-gate the provider state contract

### LINKED ISSUE #300: Functional acceptance harness: full build/destroy of a complete infra on a dedicated recette tenant

### LINKED ISSUE #306: OpenIaaS VM create hangs on an empty backup-SLA assign (state-safety: orphan-not-in-state on cancel)
## Summary

Creating an OpenIaaS virtual machine (`cloudtemple_compute_iaas_opensource_virtual_machine`) with **no `backup_sla_policies`** hangs: the resource Create unconditionally fires a backup-SLA *assign* activity — even with an **empty** policy list — and then `WaitForCompletion` blocks. On a tenant where an empty-list assign never starts platform-side, the activity stays in `Waiting` indefinitely and the apply hangs until the Create timeout.

## Impact (state-safety)

The VM is created platform-side **before** the assign step, so a hung apply that is hard-cancelled (SIGKILL) before Terraform flushes state can leave a VM that exists on the platform but is **absent from the Terraform state** — an orphan that `terraform destroy` cannot reclaim. This is exactly the state-safety class the project guards against.

## Root cause

`internal/provider/resource_compute_iaas_opensource_virtual_machine.go` (Create): the SLA assign + `WaitForCompletion` run unconditionally, even when the assembled policy list is empty:

```go
slaPolicies := []string{}
for _, policy := range d.Get("backup_sla_policies").(*schema.Set).List() { ... }
activityId, err := c.Backup().OpenIaaS().Policy().Assign(ctx, &client.BackupOpenIaasAssignPolicyRequest{
    VirtualMachineId: d.Id(),
    PolicyIds:        slaPolicies, // empty -> no-op assign that hangs
})
_, err = c.Activity().WaitForCompletion(ctx, activityId, getWaiterOptions(ctx)) // hangs
```

Assigning an empty list to a freshly-created VM has no purpose (there is nothing to clear), and the assign is only present on the Create path (Update does not re-assign).

## Reproduction (live-confirmed)

Observed on a dedicated, empty recette tenant: a `terraform apply` of an OpenIaaS VM with no `backup_sla_policies` created the VM (activity `Completed`), then the follow-up *assign job* activity stayed `Waiting for assign job activity to start` and the apply hung for 12 minutes. Passing a real SLA policy (a `nobackup` policy) instead made the assign complete and the create succeed.

## Proposed fix

- Skip the SLA assign + wait entirely when the requested policy list is empty (a fresh VM has nothing to assign/clear).
- (Already in place: `setIdFromActivityState` runs before the post-create steps, so the VM id is committed before the assign.)
- Follow-up to consider: a bounded waiter for the assign activity so a genuinely-requested-but-stuck assign surfaces an error rather than hanging.

Target: `rc/v1.8.0` (RC train). State-safety / robustness fix.

### LINKED ISSUE #316: feat(tooling): cmd/ct-validate — parametrizable endpoint validation & resilience harness

### LINKED ISSUE #318: security: TF_LOG=DEBUG can leak secretAccessKey via loggingHttpTransport
**Provider security bug.** With `TF_LOG=DEBUG`, the client's `loggingHttpTransport` logs HTTP requests/responses. The current redaction masks only HTTP **bodies**, so the Object Storage `secretAccessKey` (and any secret carried outside the masked body — a header, a URL) can appear in plaintext in the debug log.

## Impact

A user running `TF_LOG=DEBUG terraform ...` to troubleshoot can leak a long-lived object-storage secret into local logs / CI output.

## Fix

Redact sensitive fields and headers in `loggingHttpTransport` (the `Authorization` bearer token, `secretAccessKey` / `accessSecretKey`, any credential field) — not only the body. Add a test asserting the secret never appears in the logged output.

## Notes

Found during the recette harness work (#300). Secret-handling / state-safety class.

### LINKED ISSUE #327: ct-validate: preflight guards (print resolved target, refuse cleartext, require creds) before live runs
Follow-up of the #316 read-only pre-flight audit. Before the first live read-only run against the SHARED recette API, harden `cmd/ct-validate` with three operator safety guards (scoped to the tool, not the shared `internal/client`):

- **Print the resolved target** (`scheme://host`) before any request, so the operator can confirm it is the intended recette host (the default address is production-looking and the binary otherwise never shows where it fires).
- **Refuse a non-HTTPS scheme** — the credential exchange carries the PAT id/secret and must never travel in cleartext (covers both `CLOUDTEMPLE_HTTP_SCHEME=http` and an `http://` prefix in the address).
- **Fail fast on missing credentials** (incl. whitespace-only), exit 2, before the client is built or any request is sent.

Mutation-proven unit + `run()` end-to-end tests; vet / -race / staticcheck green; adversarial review GO.

### LINKED ISSUE #339: client: reads lack a per-request timeout and transient-error retry — a slow/flaky API hangs or fails plan/apply

### LINKED ISSUE #340: client: JWT() returns (nil, nil) when reading the auth response body fails — nil-token panic

### LINKED ISSUE #342: client: JWT cache-hit reads exp with unguarded type assertions — panics on a valid token without a numeric exp

### LINKED ISSUE #345: client: OpenIaaS VM create sends an empty cloudInit {} when no cloud-init is set — marketplace deploy fails 400
## Symptom

Deploying an OpenIaaS VM **from a marketplace item without any `cloud_init`** fails:

```
Error: the virtual machine could not be created from marketplace item: Unexpected response code: 400
({"errors":[{"msg":"When cloudInit is provided, cloudConfig is required","path":"cloudInit",...}]})
```

## Root cause

`CloudInit` is a **value struct** embedded with `omitempty` in both create payloads:

- `CreateOpenIaasVirtualMachineRequest.CloudInit CloudInit` (`internal/client/compute_iaas_opensource_virtual_machine.go:81`, template path)
- `MarketplaceOpenIaasDeployementRequest.CloudInit CloudInit` (`internal/client/marketplace_item.go:202`, marketplace path)

`encoding/json` `omitempty` does **not** omit a zero-value struct. The provider always builds `var cloudInit client.CloudInit` and passes it (`internal/provider/resource_compute_iaas_opensource_virtual_machine.go:449`), so a config without `cloud_init` serializes `"cloudInit":{}`. The API then requires `cloudConfig` → 400.

## Impact

Every marketplace deploy without `cloud_init` is broken (the template endpoint carries the same latent payload, though it currently tolerates the empty object). Surfaced via the `ct-validate` Terraform path (Refs #316 / #338).

## Fix

- Make the field a pointer in both request structs: `CloudInit *CloudInit \`json:"cloudInit,omitempty"\`` — a `nil` is omitted.
- Provider: build the cloud-init via a pure helper `buildOpenIaasCloudInit(raw) (*client.CloudInit, error)` that returns `nil` when `cloud_init` is absent/empty/all-empty (so `cloudInit` is omitted), an **error** when `cloud_init` is set with `network_config` but no `cloud_config` (fail closed with a clear message, instead of a raw API 400), and a populated `*CloudInit` when `cloud_config` is set.
- No read/state impact: `CloudInit` exists only in these request structs; the `cloud_init` schema attribute is unchanged (no tfstate migration).

### LINKED ISSUE #350: OpenIaaS virtual_disk: connected drifts true on a powered-off VM (perpetual reconnect plan)

### LINKED ISSUE #355: [BUG] `cloudtemple_compute_iaas_opensource_virtual_machine` — `host_id` update reports success but VM is not migrated (silent no-op)

### LINKED ISSUE #356: OpenIaaS VM: provider-initiated reboot uses the stale planned host_id instead of the live host, risking an unintended relocation

pbesret:
# Release v1.8.0

Security + bug-fix release train. **No VPC provider surface ships in v1.8.0** —
the `cloudtemple_vpc_*` datasources and resources were intentionally removed
(quarantined in #362) because the `/vpc/v1` contract is being replaced by a
breaking version under the same URL; the VPC surface is rebuilt against the new
contract in v1.9.0. The internal VPC client SDK stays quarantined and unused.

## Validation

The canonical RC validation marker (`RC-VALIDATION: OK rc/v1.8.0 <sha>`) is
posted as a PR comment per `.clinerules/WORKFLOW_GIT_EPIC.md`, once CI is green
on the final RC commit. Summary of what was validated:

- CI green on the rc/v1.8.0 tip: Build, Unit tests & vet, generate.
- Trial merge into `main` built and tested green under Go 1.24.0. The only
  conflicts were the four files touched by the `v1.7.2` hotfix (#320); `main`'s
  hotfix is fully subsumed by rc's stricter, state-safe handling, so the merged
  tree is byte-identical to the rc tip for all code (only `CHANGELOG.md`
  differs, by the v1.7.2 + v1.8.0 reconciliation).
- Full OpenIaaS compute lifecycle validated on the recette platform (create →
  converge → destroy, zero orphan).

## Conflict resolution applied in this merge

- The three OpenIaaS VM code/test files: resolved in favour of rc (superset of
  the #320 hotfix; the hotfix's two nil-safety tests already exist in rc).
- `CHANGELOG.md`: union — the `1.8.0` section on top (latest badge, dated), the
  `1.7.2` hotfix section reinserted above `1.7.1`.

## Closes

Closes #355
Closes #356
Closes #350
Closes #345
Closes #342
Closes #340
Closes #339
Closes #327
Closes #318
Closes #316
Closes #306
Closes #300
Closes #298
Closes #293
Closes #288
Closes #281
Closes #273
Closes #271
Closes #269
Closes #267
Closes #264
Closes #262
Closes #260
Closes #258
Closes #255
Closes #253
Closes #251
Closes #246
Closes #245
Closes #243
Closes #241

## Intentionally left open (do NOT close with this release)

- #238, #309, #319 — VPC work, deferred to the v1.9.0 rebuild (surface removed).
- #257 — epic, closes separately once the convergence work is fully done.


pbesret:
RC-VALIDATION: OK rc/v1.8.0 0dd11fd57ea278d8e2bf6338d969e2b023a6d1b1

## Checks performed

- **CI green** on the final RC commit `0dd11fd` (Build, Unit tests & vet, generate).
- **Conflict resolution audited.** The only merge conflicts with `main` were the four files touched by the `v1.7.2` hotfix (#320). They were resolved in favour of `rc/v1.8.0`, whose stricter, never-orphan OS-disk handling already subsumes the hotfix (same nil guards on the read loop and both flatten helpers, plus the hotfix's two nil-safety tests, already present in rc). Verified that the three OpenIaaS VM **code files are byte-identical to the pre-merge rc tip** (`b832c6e`); the only tree change introduced by this merge is `CHANGELOG.md` (1.8.0 dated + latest badge, 1.7.2 hotfix section reinserted above 1.7.1).
- **Local build/test** of the merged RC under Go 1.24.0: `go build ./...`, `go vet`, and the unit suites (`internal/provider`, `internal/provider/helpers`, `internal/client`) all pass.
- **Issue aggregation** reconstructed from the `v1.8.0` milestone (the Projects v2 board is inaccessible to the PAT). 30 delivered issues are in the `Closes` of the PR body; the VPC deferrals (#238, #309, #319) and the epic (#257) are intentionally left open.

## Upgrade risks

- **Breaking (config):** datasources `cloudtemple_compute_virtual_machine` / `cloudtemple_compute_virtual_machines` now expose `extra_config` as a **map** instead of a list of `{key, value}` objects (fixes a read that was broken since 1.6.1). Configs referencing `extra_config[0].key` must switch to `extra_config["<key>"]`.
- **Build:** the provider now requires **Go 1.24** to build from source (registry binaries are unaffected).
- **Behaviour (safer):** OpenIaaS and several other resources/datasources no longer auto-drop from state on an access-denied / inconclusive read (fail-closed, never-orphan). A previously-silent state removal now surfaces as an explicit, actionable diagnostic or as drift.
- **No VPC surface** ships in 1.8.0 (it was never published); the `cloudtemple_vpc_*` rebuild lands in 1.9.0.

## Merge instructions

`Rebase and merge` is forbidden for this PR. Use **Create a merge commit** (preserves the RC history) — `Squash and merge` is also acceptable. After merge, create the **`v1.8.0` tag** on the merge commit in `main` to trigger the release workflow.

chrlesur (changes requested):
## Validation — Release v1.8.0 (rc/v1.8.0 → main)

Independent validation + adversarial reviewer pass (Codex `gpt-5.5`, read-only,
per `.clinerules/WORKFLOW_ENGINEERING.md` §3) against the rc tip
`0dd11fd57ea278d8e2bf6338d969e2b023a6d1b1`.

**Verdict: REQUEST-CHANGES.** Three issues were found before merge. Two
(auto-close) have just been fixed in this PR body and verified; one (commit
trailers) is a merge-method decision left to the maintainer. None affects the
shipped binary — the code, CHANGELOG and #320 conflict resolution are sound.

Reviewer verdict recorded verbatim (§6): `VERDICT: NO-GO`.

### Blocking findings

**1. ✅ FIXED — issue auto-close was broken (only #355 would close).** The
`## Closes` line used a single keyword before a space-separated list
(`Closes #355 #350 …`); GitHub links only the first ref, so
`closingIssuesReferences` contained **only #355** and the rest stayed open on
merge. Corrected to one `Closes #N` per line. **Verified:**
`closingIssuesReferences` now resolves to **31 issues**, and the reverse check
`gh issue view <N> --json closedByPullRequestsReferences` returns `[372]` for
the previously-missing #350 / #258 / #293 / #241.

**2. ✅ FIXED — #356 was missing entirely.** It ships in this release (commit
`868cbdc` `fix(#356) Power OpenIaaS VM on its live host…`, documented in the
CHANGELOG) but had `milestone: null`, so the milestone-based reconstruction
dropped it. Added `Closes #356`; **verified** #356 now references PR #372.

**3. ⏳ OPEN — 7 agent-trailer commits in the release range (merge-method
decision).** `838b4b0 28f5b0a 124a7c6 b5d4642 04e4057 3475181 f4ba912` carry
`Co-Authored-By: Claude …`, which `.clinerules/WORKFLOW_ENGINEERING.md` §7
forbids "anywhere" (they entered rc via already-merged PRs #349 / #351 / #352).
A **merge commit** propagates them into `main`. Maintainer choice before merge:
- **Squash and merge** (sanctioned by the RC-VALIDATION comment) — clean single
  commit on `main`, no trailers; loses the atomic history (kept on rc).
- **Merge commit** — preserves atomic history but lands the 7 trailers on `main`.
- Rewriting rc to strip them (force-push a shared branch + re-run CI + re-issue
  RC-VALIDATION) is the most expensive and is **not** recommended.

### Non-blocking

- `VERSION` still reads `1.8.0-rc.1`. Not a release driver — goreleaser injects
  the version from the git tag (`.goreleaser.yml:20` `-X main.version={{.Version}}`);
  `main.go` defaults to `"dev"`. Cosmetic; worth bumping to `1.8.0` (or removing
  the file).
- `README.md:22` still recommends `~> 1.7`; stale once 1.8.0 publishes.
- #258 (governance/rules) is linked via milestone membership with no code commit
  in the range — defensible, just noted.

### Independently confirmed GREEN

- CI green on `0dd11fd` (Build, Unit tests & vet, generate).
- The merge `0dd11fd` has parents `b832c6e` (pre-merge rc) and `4347b98`
  (main / #320 hotfix); `git diff --name-only b832c6e 0dd11fd` returns **only
  `CHANGELOG.md`** — the merged tree is byte-identical to the rc tip everywhere
  else, so the #320 conflict was correctly resolved in favour of rc (rc already
  subsumes the hotfix; no add/add double-definition).
- CHANGELOG finalized correctly: `1.8.0 (June 26th, 2026)` + `latest` badge on
  top, `1.7.2 (June 15th, 2026)` (#320) reinserted above `1.7.1`, no stray
  `Unreleased`.
- never-orphan / fail-closed OS-disk guards intact; the hotfix's two nil-safety
  tests exist once in rc (no duplication).
- The close set is exactly `milestone-open ∖ {#238, #309, #319 (VPC→1.9.0),
  #257 (epic)}`, plus the off-milestone #356 now added. Exclusions are correct.
- No `cloudtemple_vpc_*` user surface ships; the internal VPC SDK is quarantined
  and unused by the provider; `ct-validate all` excludes the VPC cycle.
- Announced breaking changes are real and documented: datasource `extra_config`
  → `TypeMap`; Go 1.24 build baseline; fail-closed (no auto-drop on inconclusive
  read). The schema golden gate snapshot reflects `extra_config` as `TypeMap`.

### Remaining before the human merge

Only **finding 3** is left: decide the merge method (squash for a clean `main`,
or merge commit accepting the 7 trailers). Then proceed with the human merge +
the separate `v1.8.0` tag gate. The auto-close defects (1, 2) are resolved and
verified.

chrlesur (approved):
## Re-review — APPROVE

Following up on my earlier `CHANGES_REQUESTED`. All three findings are now
resolved against the rc tip `0dd11fd` (unchanged — the RC-VALIDATION marker
still holds).

- **B1 (auto-close keyword)** — fixed in the PR body (one `Closes #N` per line).
  Verified: `closingIssuesReferences` resolves to **31 issues**; the reverse
  `closedByPullRequestsReferences` returns `[372]` for the previously-missing
  ones.
- **B2 (#356 missing)** — added; verified it now references this PR.
- **B3 (7 agent-trailer commits)** — resolved by maintainer decision: merge via
  **Create a merge commit** to preserve the atomic 228-commit release history,
  consciously accepting that the 7 historical `Co-Authored-By: Claude` trailers
  (from already-merged PRs #349 / #351 / #352) land in `main`. This is a sound
  trade-off: the only way to strip the trailers would be to rewrite — i.e.
  change — the very history being preserved.

Everything else stays confirmed green (merged tree byte-identical to rc except
CHANGELOG; CHANGELOG finalized; never-orphan guards intact; VPC surface absent;
breaking changes documented; CI green).

**Merge instructions:** `Create a merge commit` (NOT rebase). After merge,
create the `v1.8.0` tag on the merge commit to trigger the release workflow.

Approving — clearing my previous change request.

Issue #384, opened 2026-06-29
client: align HTTP 403/404 handling to the API's new not-found contract


pbesret:
## Context

The API now returns `404` for an absent resource and `403` only for an
authorization failure (previously `403` was returned for both). An audit of the
provider against this contract found **no runtime regression** for existing,
readable resources: the central helper `requireNotFoundOrOK`
(`internal/client/api.go`) hardcodes `case 404` alongside the configurable
`notFoundCode`, so the absent signal migrating `403 → 404` is absorbed
identically everywhere.

However, ~56 of the 65 call sites pass `notFoundCode = 403`. Because Go matches
`switch` cases top-to-bottom, `case 404, 403:` consumes a `403` before the
dedicated `case 403:` access-denied branch is ever reached. That branch is dead
at those sites, so a genuine `403` is silently mapped to "not found".

## Consequences of the masking

No tfstate drop occurs for an existing resource (the resource read paths fail
closed, and destructive decisions rest on a strict `200`-only listing or an
authoritative `404`). But:

- **read-only List endpoints fail open**: a `403` becomes an empty "success"
  list instead of an error — e.g. `compute_snapshot`, `compute_guest_operating_system`,
  and the OpenIaaS `machine_manager` / `backup` / `policy` lists;
- the masked `403` feeds the datasource nil-pointer panics tracked separately
  in #382;
- comments across the client still assert "the API returns 403 for an absent
  resource", which is now false (documentation drift).

## Proposed change

For each endpoint whose absent response is confirmed to be `404`, switch
`requireNotFoundOrOK(resp, 403)` → `requireNotFoundOrOK(resp, 404)` so a real
`403` surfaces as an "access denied" error (aligned with the never-drop
doctrine), and update the stale comments. The VPC resource layer already
follows this model: its resolve/deprovision oracles use explicit switches that
treat `404` as the sole authoritative absence and `403` as a fail-closed error.

## Guard (important)

Do **not** flip in bulk. Confirm per endpoint, with live evidence, that an
absent resource returns `404` before switching it; an endpoint still returning
`403` for an absent resource would otherwise turn that into a hard error.
Each state-touching change goes through the project's review cycle with
RED-without-fix / GREEN-with-fix tests.

Refs #387

PR #389, opened 2026-06-29
feat(#384): read-only absence-contract probe in ct-validate

pbesret:
Refs #384

## What

Step 1 of #384: a **read-only** probe to confirm, per endpoint on a real tenant,
whether the API's new "absent → 404, 403 = forbidden only" contract is deployed —
**before** flipping any `requireNotFoundOrOK(resp, 403)` site to `404`. The flip
itself is deliberately **not** in this PR.

- `client.ProbeStatus(ctx, path, args...)` — a read-only helper that GETs `path`
  with the client's auth/base URL and returns the **raw** HTTP status, without the
  404/403 folding the resource reads apply (which would erase the very distinction
  we need). Discards the body; never mutates.
- ct-validate `probe_absence` cycle (`KindRead`, **quarantined** so it never runs
  under `-cycles all`). It GETs every by-id / parent-scoped read endpoint (the 42
  whose path has a `%s`) with a well-formed-but-nonexistent id and records the
  status: **404 → OK (migrated, safe to flip)**; **403 / other non-404 → a 4xx
  failure** (the real status in the report detail, breaker **not** tripped);
  **5xx/429 → distress** (breaker). Collection/aggregate endpoints (no `%s`) are
  excluded — a list has no "absent id".

## How to run (on recette)

```
go install ./cmd/ct-validate
ct-validate -cycles probe_absence -runs 1 -concurrency 1   # with recette creds in env
```

Reading the report: a **green/OK** probe = the endpoint returned **404** for an
absent id → safe to flip that site to `notFoundCode=404`. A **4xx** probe whose
detail shows **`HTTP 403`** = **not migrated** → do **not** flip it.

**Operational caveat:** a 403 is an unambiguous "not migrated" signal only with a
token that **has read permission** on that resource type (otherwise a 403 could be
a genuine permission denial). Run with a broadly read-entitled recette token.

## Tests (mutation-proven)

- `client`: `ProbeStatus` returns the raw status (200/403/404/409).
- `ct-validate`: the status→outcome mapping (404=OK, 403/anomalies=non-distress
  4xx, 5xx/429=distress); the cycle records correctly and the breaker never trips
  on 403s; an **exact registry-coverage** test scans the client and fails if a
  `403`-by-id endpoint has no probe entry (or a stale entry lingers); and the cycle
  is quarantined yet runnable when named.
- Mutation-proven: a 404→error makes the mapping test RED; dropping a registry
  entry makes the coverage test RED.

## Review

Independent adversarial review by Codex (read-only): PLAN review GO (with 3
refinements, all applied), committed-diff review GO (no blocking findings).

Internal test-harness tooling (`cmd/ct-validate`), not a provider-user-facing
change → no CHANGELOG entry. `Refs #384` (no closing keyword, RC flow).

### DIFF
--- a/cmd/ct-validate/cycle_probe_absence.go
+++ b/cmd/ct-validate/cycle_probe_absence.go
@@ -23,10 +23,13 @@ type probeAbsenceEndpoint struct {
 // UUID shape for id routes, valid object-storage names, a valid MAC for ReadByMAC,
 // and a real marketplace provider target for the /items/%s/%s/info route.
 const (
-	probeBogusID     = "00000000-0000-0000-0000-000000000000"
-	probeBogusID2    = "11111111-1111-1111-1111-111111111111"
+	// Random-looking valid UUIDs (NOT the nil/all-ones UUID: a stricter route
+	// validator may special-case those and answer 405/400 instead of running the
+	// absence lookup).
+	probeBogusID     = "7f3e9a2b-1c4d-4e5f-8a6b-9c0d1e2f3a4b"
+	probeBogusID2    = "b2c3d4e5-6f70-4a1b-9c2d-3e4f5a6b7c8d"
 	probeBogusName   = "ctvalidateprobeabsent"
-	probeBogusMAC    = "00:00:00:00:00:00"
+	probeBogusMAC    = "02:00:5e:10:00:01"
 	probeMarketplace = "vmware"
 )
 
--- a/cmd/ct-validate/cycle_probe_absence.go
+++ b/cmd/ct-validate/cycle_probe_absence.go
@@ -1,0 +1,158 @@
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/cloud-temple/terraform-provider-cloudtemple/internal/client"
+)
+
+// probeAbsenceEndpoint is one by-id / parent-scoped read endpoint to probe with a
+// deliberately-nonexistent id, to observe whether the API returns 404 (the new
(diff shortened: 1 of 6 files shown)


pbesret:
**Disposition (agreed 2026-07-03): release-gate tool, not code to ship.**

This PR is parked until **Compute API v1.144.0 is deployed in production**. The 403→404 flip it was written to validate is already merged in `rc/v1.9.0` (hence the Compute ≥ 1.144.0 requirement in the v1.9.0 UPGRADE NOTES).

**v1.9.0 release runbook — step 0:**
1. Build `ct-validate` **from this branch** and run `ct-validate -cycles probe_absence -runs 1 -concurrency 1` against production with a broadly **read-entitled** token (read-only cycle; a 403 is only meaningful with read permission).
2. All 42 by-id endpoints report **404 for an absent id** → the v1.9.0 contract is confirmed safe → proceed with the release and **close this PR without merging** (the probe has served its purpose).
3. Any endpoint still answering **403** → do not release; report the endpoint upstream.

Alexandre-Cann:
Closing without merge. The probe_absence cycle served its purpose as the v1.9.0 release gate: run live against production it confirmed the 404-on-absent contract on all 29 Compute endpoints (17 VMware + 12 OpenIaaS), zero 403 — prod Compute API = v1.144.1 (>= v1.144.0). The diagnostic is intentionally not merged; v1.9.0 has been tagged (main @ e457ea0) and the branch feat/384-absence-probe is being deleted.

Issue #382, opened 2026-06-29
provider: datasource reads can panic (nil pointer) instead of returning a clean not-found/forbidden error


pbesret:
## Summary

Several datasource reads dereference the value returned by a client read method
without first checking it for `nil`. The client read maps an HTTP not-found or
forbidden response to `(nil, nil)` (through `requireNotFoundOrOK`), so when the
target resource is absent — or when the caller's token lacks read permission —
the datasource panics instead of returning a clean diagnostic.

## Affected sites

- `internal/provider/data_source_object_storage_bucket.go:115` —
  `d.SetId(bucket.ID)` runs right after `c.ObjectStorage().Bucket().Read(...)`
  with no `bucket == nil` guard; `Read` returns `(nil, nil)` for an absent or
  forbidden bucket (`requireNotFoundOrOK(resp, 403)`), so `bucket.ID` panics.
- `internal/provider/data_source_backup_metrics.go` (`history` ~line 262,
  `platform_cpu` ~line 268) — `helpers.FlattenBackupMetricsHistory` and
  `FlattenBackupMetricsPlatformCPU` dereference a `nil` pointer when
  `Metrics().History(...)` / `Metrics().PlatformCPU(...)` return `(nil, nil)`
  (`requireNotFoundOrOK(resp, 403)` at `internal/client/backup_metrics.go:33`
  and `:164`). The `err` check passes (`err` is `nil`), so the read falls
  through to the flatten helper.

## Trigger

A non-existent name/id, or a token lacking read permission on the resource.
With the API now returning `404` for an absent resource and `403` only for an
authorization failure, the forbidden path is a legitimate, reachable runtime
condition — which makes these crashes more likely to be hit in practice.

## Expected behavior

Return a clear diagnostic (for example `"<resource> not found"` or
`"access denied"`) instead of panicking. A datasource read that finds nothing
to read must fail with an actionable error, never crash the provider.

## Notes

- Pre-existing defect, independent of the 403→404 status-code change: the crash
  occurs identically whether the absent signal arrives as `403` or `404`.
- It does not affect the tfstate of an existing, readable resource.
- The fix should land on the current RC branch (`rc/v1.9.0`).

Refs #387

PR #486, opened 2026-07-07
fix(object-storage): éviter le nil-panic des data sources bucket & storage account sur 404

KChapron:
## Contexte

La campagne de normalisation des status code Shiva fait renvoyer **HTTP 404** (au lieu de 500) par l'API object-storage quand un bucket ou un storage account est **absent**.

Côté provider, le client mappe 404 → `(nil, nil)` via `requireNotFoundOrOK`. Ensuite `dataSourceBucketRead` et `dataSourceStorageAccountRead` déréférençaient le pointeur nil (`d.SetId(bucket.ID)` / `account.ID`) → **panic au refresh** (avant, le 500 produisait une `diag` propre).

## Correctif

Les deux lectures de data source échouent désormais proprement avec un diagnostic actionnable quand l'objet est introuvable, au lieu de paniquer :

```go
if bucket == nil {
    return diag.Errorf("no bucket found with name %q", bucketName)
}
```
(idem `account == nil` dans `data_source_object_storage_storage_account.go`)

Ce sont les **seules** data sources object-storage avec le pattern `Read()` → `d.SetId(x.ID)` (vérifié par grep) ; les listes (`buckets`, `storage_accounts`, `role`) renvoient des slices et ne sont pas concernées.

## Vérification
- `gofmt` : clean
- `go build ./internal/provider/` : OK

## Backport
À rétroporter sur **`rc/v1.9.0`** et **`rc/v1.10.0`** (les data sources y sont identiques).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/provider/data_source_object_storage_bucket.go
+++ b/internal/provider/data_source_object_storage_bucket.go
@@ -110,6 +110,12 @@ func dataSourceBucketRead(ctx context.Context, d *schema.ResourceData, meta any)
 	if err != nil {
 		return diag.FromErr(fmt.Errorf("error reading bucket with name %s: %s", bucketName, err))
 	}
+	// The client maps a not-found (HTTP 404) / access-denied answer to (nil, nil).
+	// A data source must resolve, so fail closed with an actionable diagnostic
+	// instead of dereferencing a nil bucket (which would panic).
+	if bucket == nil {
+		return diag.Errorf("no bucket found with name %q", bucketName)
+	}
 
 	// Définir l'ID de la datasource
 	d.SetId(bucket.ID)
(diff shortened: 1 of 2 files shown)


pbesret (commented):
**Review — superseded by #385 (already released)**

Thanks for this fix — the analysis is correct: the client maps a 404/403 answer to `(nil, nil)` via `requireNotFoundOrOK`, and both datasource reads used to dereference the nil pointer.

However, `main` already carries the equivalent guard since 3ad5bde (fix(#382), PR #385, merged June 29th), released in v1.9.0 and v1.10.0: both `dataSourceBucketRead` and `dataSourceStorageAccountRead` fail closed with an actionable diagnostic covering both the not-found and access-denied cases, and the guards are covered by unit tests (`internal/provider/datasource_nil_read_unit_test.go`: 404/403 nil-reads + success paths). That pre-existing guard is also why this PR shows conflicts: both branches add a guard at the same location in the same two files.

Checks performed: compared this diff against main's 3ad5bde; ran the object-storage datasource unit tests on main (all pass); verified the guard ships in v1.9.0+. An independent adversarial review (Codex) confirmed the conclusion, noting additionally that main's diagnostic is more precise for the 403 case ("does not exist or access is denied" versus "no bucket found").

Conclusion: no remaining gap for this PR to fill — I suggest closing it as superseded by #385. (Informative review, no change requested.)

Issue #393, opened 2026-06-29
compute_iaas_opensource_virtual_machine: acceptance tests use ignore_changes=[memory] and commented assertions — complacent test pattern (§4 violation)


chrlesur:
Multiple acceptance tests touching `memory` in the OpenIaaS resource carry `lifecycle { ignore_changes = [memory] }` AND have the corresponding `TestCheckResourceAttr` assertions on `memory` commented out. As a result, if `Update` silently no-op'd any memory change, these tests would stay green — a direct violation of `WORKFLOW_ENGINEERING.md` §4 "no complacent or vacuous tests".

## Occurrences (provider main @ v1.8.0)

- `internal/provider/tests/resource_compute_iaas_opensource_virtual_machine_test.go` lines 47, 61, 110, 142
- `internal/provider/tests/resource_compute_iaas_opensource_virtual_machine_migration_test.go` line 81
- Equivalent patterns in adjacent disk and network tests around `memory`

## Why this matters

`WORKFLOW_ENGINEERING.md` §4 requires every fix or guard to ship with a test that goes RED without it. The current memory tests cannot — they ignore the attribute they should pin. If the silent-drift bug suspected in #392 is real, the existing test suite would not catch it.

## Proposed fix

- Remove `lifecycle { ignore_changes = [memory] }` from these test cases.
- Re-enable the `TestCheckResourceAttr(..., "memory", "<expected>")` assertions.
- Add a TestStep that exercises Update memory and asserts the API GET (via a TestCheckFunc that calls the client directly) matches the HCL value post-apply — mutation-proof: RED if Update is silently no-op'd, GREEN otherwise.

## Refs

- Parent investigation: #392
- `.clinerules/WORKFLOW_ENGINEERING.md` §4 test discipline
- Epic: #387 (v1.9.0 hardening)

Issue #391, opened 2026-06-29
platform/api: /backup/v1/open_iaas/policies intermittent multi-minute hangs (root cause behind #339; OpenIaaS counterpart of #315)


chrlesur:
Platform reliability issue — OpenIaaS counterpart of #315 (VPC ComputeManager) on the backup-policies read path.

## Symptom

`GET /api/backup/v1/open_iaas/policies` intermittently hangs for several minutes before returning normally — the exact symptom that prompted #339 ("Observed live: a `GET /backup/v1/open_iaas/policies` hung ~3 minutes during a datasource read before manual abort").

## Latest live reproduction (2026-06-29)

- Recette OpenIaaS, `shiva.cloud-temple.com`, tenant `fc37d401-…`.
- During a `terraform apply` (provider built from `main` @ v1.8.0):
  - `data.cloudtemple_compute_iaas_opensource_availability_zones` stalled ~2 min.
  - `data.cloudtemple_backup_iaas_opensource_policies` was still "Still reading… [03m50s elapsed]" when the run was manually aborted.
- ~30 minutes later, a direct `curl` to the same endpoint returned **HTTP 200 in 306 ms** (5 policies, 1577 bytes; TLS 74 ms, TTFB 304 ms).
- Cross-check `GET /api/backup/v1/spp/policies` (VMware SPP equivalent, same proxy chain): 132 ms — slowness is isolated to the OpenIaaS backup path, not the HTTP/IAM/proxy stack.
- Therefore: a transient platform-side condition (likely a lock or queue on the OpenIaaS / Xen Orchestra policy backend) sporadically inflates this endpoint's latency by ~750× over normal.

## Relationship with #339 and #315

- **#339** (CLOSED 2026-06-26 via #372, v1.8.0) ships the **provider-side** mitigation: `CLOUDTEMPLE_HTTP_TIMEOUT` (default 240 s) plus a bounded retry on idempotent GETs. It turns an infinite hang into either a transient retry success or an actionable failure. It does **not** fix the platform-side cause — #339 explicitly scopes that out:
  > Platform-side flakiness on ComputeManager is tracked in #315; this issue is the provider-side resilience that should hold regardless of the API's behaviour.
[...]

PR #487, opened 2026-07-08
Release v1.9.0 — VPC integration + provider hardening

### LINKED ISSUE #238: VPC provider surface: datasources, resources, OpenIaaS association

### LINKED ISSUE #374: feat(vpc): cloudtemple_compute_iaas_opensource_network_adapter — add ip_address to attach a VM to a VPC private network (OpenIaaS)
Refs #238. **Mandatory for v1.9.0.** Without this, a VM network adapter cannot be placed on a VPC private network with a controlled static IP — the core capability v1.9.0 must deliver. The `cloudtemple_vpc_floating_ip_binding` resource does NOT cover this (it binds a floating IP to a static IP; it is unrelated to attaching an adapter to a VPC network).

## Problem

The provider sends only `{ virtualMachineId, networkId, mac }` on create and `{ networkId, mac, attached, txChecksumming }` on update of an OpenIaaS network adapter. To attach an adapter to a VPC-backed network **with a chosen static IP**, the Compute API needs an `ipAddress` field the provider does not send.

## API contract (authoritative)

Compute broker commit `c1b6f461e` — `feat(#1854): add VPC conf on adapter update and creation` (2025-10-22) added `ipAddress` (IPv4) to the create **and** update schemas of the OpenIaaS network-adapter endpoints. Server logic (`XOANetworkAdapterController` → `VpcManager.registerStaticIp`): when the target network has a `private_network_id` (i.e. it is VPC-backed), the adapter create/update registers a `source: xoa` static IP on the private network for the adapter's MAC — at `ipAddress` when provided, otherwise auto-assigned.

## Live evidence (tenant FW_AGENTIC, shiva.cloud-temple.com, 2026-06-27)

- `POST /compute/v1/open_iaas/network_adapters { virtualMachineId, networkId=<VPC-backed>, ipAddress: 192.168.192.51 }` → 201 + activity completed → a static IP appears in `GET /vpc/v1/private_networks/{pn}/static_ips` at **exactly 192.168.192.51**, `source: xoa`, linked to `{ networkAdapter.id, macAddress, virtualMachine }`.
- The same call **without** `ipAddress` → static IP auto-assigned (first free, e.g. 192.168.192.10), `source: xoa`.
- `DELETE` the adapter → the static IP is released (zero orphan — the #311/#359 platform fix, reconfirmed live).

## Implementation scope

- **Client**: add `IPAddress string \`json:"ipAddress,omitempty"\`` to `CreateOpenIaasNetworkAdapterRequest` and `UpdateOpenIaasNetworkAdapterRequest`.
- **Resource** `cloudtemple_compute_iaas_opensource_network_adapter`: add `ip_address` (Optional, IPv4 `ValidateFunc`). Plumb on create and update.
- **Read / convergence (important)**: the adapter read does NOT echo the assigned static IP — `vpc.staticIpAddress` is absent when the VM has no live guest IP (e.g. halted, verified live). To read back / converge `ip_address`, query `GET /vpc/v1/private_networks/{pn}/static_ips` and match by MAC (or `networkAdapter.id`). Define the convergence model so a configured `ip_address` does not perpetually diff.
- **Doctrine (validated with the product owner)**: `ip_address` on the adapter is the **canonical** VPC-addressing path (registers a `source: xoa` static IP); `cloudtemple_vpc_static_ip` stays for `source: custom` off-adapter allocations. Document the split; guard against double-managing the same MAC on the same private network.
- **State-safety**: only meaningful on a VPC-backed network; a failed VPC registration surfaces server-side as "VPC configuration incomplete" → fail the apply with a clear diagnostic, never a silent partial. Delete releases the static IP (verified).
- **Tests**: non-complacent unit tests + live E2E validation before merge.

## Acceptance criteria

- `ip_address` settable on create and update; the VPC static IP lands at the chosen address (verified live).
- Plan converges (no perpetual diff) both with and without `ip_address`.
- Destroy leaves zero orphaned static IP.

### LINKED ISSUE #375: feat(vpc): cloudtemple_compute_network_adapter — add ip_address for VPC attachment + first live VMware↔VPC validation
Refs #238. **Mandatory for v1.9.0.** VMware counterpart of #A: a `cloudtemple_compute_network_adapter` cannot today be attached to a VPC private network with a controlled static IP.

## Problem

The VMware adapter resource sends `{ virtualMachineId, networkId, type, macAddress }` (create) and `{ id, newNetworkId, autoConnect, macAddress }` (update), and its schema has **no VPC fields at all**. To attach to a VPC-backed network with a chosen static IP, the Compute API needs `ipAddress`, which the provider does not send.

## API contract (authoritative)

The same commit `c1b6f461e` — `feat(#1854): add VPC conf on adapter update and creation` (2025-10-22) added `ipAddress` (IPv4) to the **vCenter** network-adapter create and update schemas (`vcenters/networkAdapters`), and `VCenterController.createNetworkAdapter` / `updateNetworkAdapter` forward it. The same `VpcManager.registerStaticIp` path applies when the target vCenter network is VPC-backed (`private_network_id` set), with `source: vmware`.

## Important: VMware↔VPC is unproven live

Per #238 "Known gap", the VMware↔VPC association is *structurally supported by the model but not yet validated end-to-end on a live VMware VM*. The OpenIaaS path is now proven live (see #A); **the VMware path is not**. This issue MUST include the **first live VMware↔VPC validation** on a vCenter tenant, not only the code change.

## Implementation scope

- **Client**: add `IPAddress string \`json:"ipAddress,omitempty"\`` to `CreateNetworkAdapterRequest` and `UpdateNetworkAdapterRequest`.
- **Resource** `cloudtemple_compute_network_adapter`: add `ip_address` (Optional, IPv4 `ValidateFunc`). Plumb on create and update.
- **Read / convergence**: confirm on vCenter whether the assigned static IP is readable from the adapter or only via `GET /vpc/v1/private_networks/{pn}/static_ips` (matched by MAC) — likely the latter, as on OpenIaaS. Define the convergence model accordingly.
- **Doctrine**: same as #A — `ip_address` on the adapter = canonical VPC addressing (`source: vmware`); `cloudtemple_vpc_static_ip` = custom off-adapter.
- **State-safety**: fail-closed on a failed VPC registration; destroy must release the static IP (verify live, given the #311/#359 history was observed on OpenIaaS — confirm VMware behaves the same).
- **Tests**: non-complacent unit tests + **mandatory first live VMware↔VPC E2E** before merge.

## Acceptance criteria

- `ip_address` settable on create and update on a VMware adapter; the VPC static IP lands at the chosen address — **verified live on a vCenter tenant**.
- Plan converges with and without `ip_address`.
- Destroy leaves zero orphaned static IP (verified live).

### LINKED ISSUE #379: refactor: unify the shared VPC network-adapter ip_address helpers (OpenIaaS + VMware)
Follow-up cleanup after #374 (PR #377, OpenIaaS) and #375 (PR #378, VMware) land. Mandatory before the v1.9.0 release: no duplicated logic should ship.

## Context

Both `ip_address` implementations needed the same two pure helpers. Because #374 and #375 are independent branches/PRs targeting `rc/v1.9.0`, defining the same package-level symbol on both would collide at merge, so #375 carries temporary VMware-prefixed **copies** (flagged in-code with "mirrors the OpenIaaS … (#374); to be unified once both land"):

| #374 (OpenIaaS) | #375 (VMware) — duplicate |
|---|---|
| `vpcStaticIPToPush` | `vmwareVPCStaticIPToPush` (identical) |
| `adapterVPCStaticIP` | `vmwareAdapterVPCStaticIP` (identical) |
| `ensureVPCForIPAddress` | `ensureVMwareVPCForIPAddress` (**differs**: reads the vCenter vs OpenIaaS network) |

Both resources are in the same `provider` package, so once #374 is on `rc` the dedup is straightforward.

## Plan (after both PRs merge to `rc/v1.9.0`)

1. Move the canonical pure pair (`vpcStaticIPToPush`, `adapterVPCStaticIP`) into a neutral file, e.g. `internal/provider/vpc_network_adapter_ip.go` (not "owned" by the OpenIaaS resource).
2. Have the VMware resource call them; **delete** `vmwareVPCStaticIPToPush` / `vmwareAdapterVPCStaticIP` and their now-redundant unit tests (the shared pair stays covered by the existing matrix).
3. Optionally factor `ensureVPCForIPAddress` / `ensureVMwareVPCForIPAddress` into one helper parameterised by a network-VPC-check callback (they only differ in the network read).
4. Re-run the CI bar (vet, ratchet, `go generate`) + a Codex pass; the golden snapshot must not change (pure refactor).

## Acceptance

- No duplicated helper logic across the two adapter resources.
- Schema golden snapshot unchanged (behaviour-preserving refactor).

Depends on: #374 (PR #377) and #375 (PR #378).

### LINKED ISSUE #382: provider: datasource reads can panic (nil pointer) instead of returning a clean not-found/forbidden error

### LINKED ISSUE #384: client: align HTTP 403/404 handling to the API's new not-found contract

### LINKED ISSUE #386: provider: VMware virtual_machine resource operations can panic (nil-deref) on a nil VM read
## Summary

The VMware `cloudtemple_compute_virtual_machine` resource dereferences the value
returned by `Compute().VirtualMachine().Read` without a nil check in several
operation paths. That client read maps an HTTP 404/403 response to a `(nil, nil)`
result via `requireNotFoundOrOK(resp, 403)` (`internal/client/compute_virtual_machine.go:229`),
so an operation on a VM that is absent — or that the token is not allowed to read
— can panic (nil-pointer dereference) instead of returning a clean error.

## Affected sites

- `internal/provider/resource_compute_virtual_machine.go:1348` — Update, `customize` change
- `internal/provider/resource_compute_virtual_machine.go:1557` — Update, power-state change
- `internal/provider/resource_compute_virtual_machine.go:1588` — Delete

Each does `vm, err := c.Compute().VirtualMachine().Read(ctx, d.Id())`, checks only
`err`, then dereferences `vm`.

## Context

This is the same defect class as #382 (datasource nil-deref), found by the
independent reviewer during the #382 committed-diff review, but here in the
**resource** operation paths rather than datasources. The VM resource's
drift-Read path is not affected (it already fails closed through
`confirmVMwareDeviceOrKeep`).

## Proposed work

- Add a nil guard at each of the three sites returning an actionable diagnostic
  (the operation cannot proceed on a VM that cannot be read).
- Do a repo-wide sweep for the same `(nil, nil)`-then-dereference pattern: every
  `requireNotFoundOrOK(resp, 403)` client read whose result a caller dereferences
  without a nil check, so the closure is complete rather than site-by-site.
- Mutation-proven unit tests, as in #382.

Independent of the `403 → 404` client-mapping alignment (#384). The fix should
land on the current RC branch (`rc/v1.9.0`).

Refs #387

### LINKED ISSUE #387: Epic - Terraform Provider Cloud Temple : v1.9.0 — VPC integration + provider hardening
# Epic - Terraform Provider Cloud Temple : v1.9.0 — VPC integration + provider hardening

## Why this epic

This epic carries the v1.9.0 delivery train of the Cloud Temple Terraform
provider. It continues the "API convergence and state safety" initiative started
in #257 — which tracked the v1.8.0 convergence work and is now delivered and
released — and scopes the remaining and new work that ships with v1.9.0.

As in #257, the primary constraint is Terraform state safety: schema, read,
import, refresh, update and delete changes must not corrupt existing client
states, create permanent drift, or trigger destructive plans. Internal support
details, tenant-specific evidence and raw logs are tracked outside this public
issue.

Predecessor: #257 (v1.8.0 — API convergence & state safety), delivered and
released; closed and kept as the v1.8.0 record.

## Scope

In scope (v1.9.0):

- VPC integration: datasources and reversible write resources, and the
  dependency on the platform Order API for full CRUD.
- Provider hardening around the API's HTTP not-found contract: the API now
  returns 404 for an absent resource and 403 only for an authorization failure
  (previously 403 for both). Align the provider's client status mapping and fix
  the latent nil-dereference crashes surfaced by that audit, without changing the
  never-drop / never-orphan doctrine.

Out of scope:

- Tenant-specific incident details, internal support or operations traces.
- `cloud_init` ephemeral-resource compatibility (#237) — tracked separately in
  the backlog, not part of the v1.9.0 train.

### VPC integration

- #238 — VPC provider surface (datasources + reversible write resources;
  full CRUD for `cloudtemple_vpc` / `cloudtemple_vpc_private_network` blocked on
  the Order API).
- #308 — `cloudtemple_vpc` + `cloudtemple_vpc_private_network` resources via the
  Order API (blocked on missing platform endpoints).

### HTTP 403/404 not-found contract — hardening

Outcome of the 403→404 audit (no runtime regression found; the absent signal is
absorbed identically, but a few latent issues are worth closing):

- #382 — datasource reads can panic (nil pointer) on an absent/forbidden read.
  Fixed; merged to `rc/v1.9.0`, awaiting release.
- ~~#383~~ — `guest_operating_system` acceptance test: closed as won't-fix (not part of the v1.9.0 train).
- #384 — align HTTP 403/404 handling to the API's new not-found contract
  (umbrella). Delivered as two flips merged to `rc/v1.9.0`: #390 (Backup,
  Marketplace and VPC reads) and #404 (Compute reads, VMware + OpenIaaS). A
  genuine 403 now surfaces as an access-denied error instead of being masked as
  not-found; only a definitive 404 is treated as absent. The Compute flip (#404)
  carries a deployment dependency — see Release gate below.
- #386 — VMware `virtual_machine` resource operations can panic (nil-deref) on a
  nil VM read.

### Release gate (v1.9.0)

- ⚠️ **Do NOT tag/release v1.9.0 until the production Compute API is >= v1.144.0**
  (the first version returning 404-on-absent for compute). The #404 Compute flip
  relies on that contract: against an older Compute API (which returns 403 for an
  absent resource), an absent or deleted compute resource would surface as an
  "access denied" error instead of being treated as removed — a loud failure,
  never silent state loss. Verify the deployed Compute API version (e.g. re-run
  the ct-validate absence probe to confirm 404-on-absent) before cutting the
  RC → `main` release. Backup/Marketplace/VPC (#390) already honour the 404
  contract. Documented in the CHANGELOG `UPGRADE NOTES`.

## Project tracking

Operational tracking (prioritization, risk classification, owners, internal
API/support items, state-safety tracking) is done in the private Cloud-Temple
organization Project. This public issue is the public anchor for the v1.9.0
provider train. Delivery follows the Release Candidate flow on `rc/v1.9.0`;
issues close at the RC → `main` release.

### LINKED ISSUE #394: compute_virtual_machine: os_disk.capacity Update path lacks HasChange + pre-flight shrink check; opaque "Invalid operation" activity on shrink
Side observation while investigating #392: provisioning a VMware VM from a marketplace template (Debian 12, 32 GiB OS disk default) with HCL `os_disk { capacity = 10737418240 }` triggers a separate `Update virtual disk` activity on the marketplace-provided disk. The API refuses the activity:

> failed to update virtual disk, ... activity state: failed
> Reason: Invalid operation for device '0'.

The schema docstring confirms the constraint: `capacity: "The size of the disk in bytes. The size must be greater than or equal to the size of the virtual machine's operating system disk."` But the validation lives entirely on the platform: the provider always tries to PATCH.

After the failed apply, `terraform state show` reports the disk's API-reported `capacity` (32 GiB, the actual disk on the platform), but if the user explicitly sets a smaller capacity, the conflicting value re-triggers the PATCH activity on every plan/apply cycle. The `Optional+Computed` flag (`resource_compute_virtual_machine.go:405-409`) mitigates the worst case (the field is absorbed from the API rather than re-asserted when unset), but the underlying Update code does NOT gate the PATCH on `d.HasChange("os_disk")`, so a user who explicitly sets a smaller capacity will see the same activity rejection on every cycle.

This is the same class of issue as `memory` (#392): Update sends fields whose effects the platform may silently refuse, and the provider does not validate the user's intent against feasible API operations before calling.

## Why this matters

`.clinerules/WORKFLOW_ENGINEERING.md` §5: any destructive or refuse-able decision must come from declared input + external authority, with fail-closed when in doubt. Today an explicit capacity shrink is silently retried until the user notices.

## Proposed fix

1. Add `d.HasChange("os_disk")` gating in Update — only PATCH the disk if the user actually changed something.
2. Add a pre-flight check: if the user-specified `capacity` is less than the current API-reported capacity, return `diag.Errorf("cannot shrink an existing OS disk from X to Y; vSphere refuses this operation")` BEFORE calling the API.
3. Promote the schema's docstring constraint into a `ValidateFunc` if possible (compare with API-reported size at plan-time).

Test: an apply that lowers `os_disk.capacity` below the current value should fail fast with the diag, not via an opaque "Invalid operation" activity.

## Refs

- Parent investigation: #392
- `.clinerules/WORKFLOW_ENGINEERING.md` §5 state-safety
- Epic: #387 (v1.9.0 hardening)

### LINKED ISSUE #395: compute_virtual_machine: memory schema Default 32 MiB silently proposes a shrink when HCL omits memory (§5 violation)
`internal/provider/resource_compute_virtual_machine.go:329` declares:

```go
"memory": {
    Type:        schema.TypeInt,
    Description: "In bytes. The quantity of memory to start the virtual machine with.",
    Optional:    true,
    Default:     33554432,
},
```

The Terraform SDK cannot distinguish "user omitted `memory`" from "user set `memory = 33554432`" — both produce the same plan input. With this Default + `Optional: true` (and no `Computed: true`), if a user removes `memory` from their HCL or never sets it, Terraform plans a resize to **32 MiB** silently. For a VM that runs in production with N GiB of RAM, this is a class-level state-safety violation.

## Why this matters

`.clinerules/WORKFLOW_ENGINEERING.md` §5: "Intent comes from the declared input compared to the external authority — not from defaults or zero values the framework cannot distinguish from an unset field." A `Default` whose interpretation collapses unset-vs-set is exactly what §5 forbids.

## Behavioural reproduction

Side observation while investigating #392: on a Debian 12 marketplace VM (32 GiB OS disk default), HCL `os_disk { capacity = 10737418240 }` triggered an `Update virtual disk` activity refused by the API. The same class of mistake on `memory` would propose a quiet `Default → 32 MiB` plan that an inattentive user could approve, since the diff would just be `+ memory = 33554432` and Terraform would happily `apply`.

## Proposed fix

Two viable options:
1. Drop `Default: 33554432` and make `memory` `Required: true` (cleaner — every user must declare an explicit value).
2. Promote to `Computed: true, Optional: true` without `Default` — the value is then absorbed from the API at create-time and stays unless the user explicitly overrides.

Option 1 is safer (§5 favours explicit intent). Option 2 is more flexible.

A mutation-proof test must accompany the change: HCL without `memory` after import should NOT plan a 32 MiB resize.

## Refs

- Parent investigation: #392
- `.clinerules/WORKFLOW_ENGINEERING.md` §5 state-safety
- Epic: #387 (v1.9.0 hardening)

### LINKED ISSUE #396: compute_iaas_opensource_virtual_machine: doctrine #267 "PATCH only diffs" can short-circuit memory Update without HasChange and without read-back
`internal/provider/resource_compute_iaas_opensource_virtual_machine.go:1285-1289`:

```go
if desired.Memory != 0 && desired.Memory != live.Memory {
    req.Memory = desired.Memory
    changed = true
    needsReboot = true
}
```

Issue #267 introduced "PATCH only diffs": before sending an Update, the provider GETs the live state and compares each field against the desired (HCL) value. Only diverging fields are PATCHed. This is a sound optimisation in steady-state, but it creates a state-safety hole when:

1. A previous Apply N pushed `memory=4GiB` to the API.
2. The activity returned `completed` (the platform acknowledged the configuration update).
3. The XCP-ng materialisation lags or fails silently (the running VM is still on 2 GiB).
4. Apply N+1 (or simply a `terraform plan` cycle) sees `live.Memory == desired.Memory == 4 GiB` → the if-guard short-circuits → no PATCH → no re-attempt of the orchestrated power-off/PATCH/power-on dance.

Two compounding factors:
- The guard does NOT check `d.HasChange("memory")` — it only compares live vs desired. If the live was forged by a prior partial Apply, the comparison is fooled.
- There is no read-back assertion after `WaitForCompletion(activityId)`: the provider trusts the platform's activity acknowledgement as the only source of truth about whether the change materialised.

This is the **most likely** structural explanation for the symptom reported by the client in #392, although we could not reproduce it live on recette OpenIaaS (the platform always materialised our PATCHes promptly in the tested window).

## Why this matters

`.clinerules/WORKFLOW_ENGINEERING.md` §5: destructive (or in this case, no-op-by-comparison) decisions must come from declared input vs the external authority. Today a doctrine intended to reduce API churn can silently skip the operation the user explicitly asked for.

## Proposed fix

1. Add `d.HasChange("memory")` (and other fields touched by #267 logic) as an ADDITIONAL precondition to skip the PATCH:
   ```go
   if d.HasChange("memory") && desired.Memory != 0 && desired.Memory != live.Memory {
       ...
   }
   ```
   With this, when the user explicitly changes memory in HCL, the PATCH is sent regardless of what `live` says.
2. Add a read-back assertion AFTER `WaitForCompletion`: GET the VM, compare `vm.Memory` to `req.Memory`, and `diag.Errorf` if they diverge — fail-closed instead of silent success.
3. Document the doctrine #267 limitation in code comments at the guard site.

## Mutation-proof test

Simulate Apply N+1 where the in-memory `live` was manipulated to equal `desired` but the underlying VM still has the old value. Expect the PATCH to be sent because `d.HasChange("memory")` is true.

## Refs

- Parent investigation: #392
- Doctrine: #267
- `.clinerules/WORKFLOW_ENGINEERING.md` §5 state-safety
- Epic: #387 (v1.9.0 hardening)

### LINKED ISSUE #397: compute_virtual_machine: memory Update sends Ram unconditionally; explicit error on hot-add OFF + power=on; need allow_vm_restart provider-side attribute
The VMware VM Update path in `internal/provider/resource_compute_virtual_machine.go:1274-1298` builds the API payload with `Ram: d.Get("memory").(int)` (and `Cpu`, `CorePerSocket`, `HotCpuAdd`, etc.) and PATCHes the VM unconditionally, regardless of which fields actually changed:

```go
req := &client.UpdateVirtualMachineRequest{
    Id:                           d.Id(),
    Ram:                          d.Get("memory").(int),
    MemoryReservation:            d.Get("memory_reservation").(int),
    Cpu:                          d.Get("cpu").(int),
    ...
}
```

There is no `d.HasChange("memory")` gating, no pre-power-off conditional, and no opt-in restart attribute. When `memory_hot_add_enabled=false` and `power_state="on"`, the API rejects the activity with an explicit error:

> failed to update virtual machine, an error occured while waiting for completion of activity ...
> Reason: Memory hot plug is not supported for this virtual machine.

(Verified live during the investigation behind #392.)

Worse: between the failed apply and a subsequent `terraform refresh`, the TF state file holds the **intended** target memory value (e.g. 4 GiB) while the API still reports the previous value (2 GiB). A user reading `terraform state show` between the failed apply and the next refresh sees the desired-but-not-materialised value.

## Why this matters

`.clinerules/WORKFLOW_ENGINEERING.md` §5 state-safety: the destructive decision (forcing a PATCH the API will reject if hot-add is off) must be conditional on the declared intent, not unconditional. The current code path:

1. Hits the API on every Update.
2. Returns a failure that the user must interpret without provider help.
3. Leaves the TF state in an intermediate inconsistent shape until the next refresh.

The VMware API has NO `allow_vm_restart` / `forceRestart` equivalent flag (confirmed via swagger.comput.yml audit). The opt-in restart must therefore be a **provider-side attribute**.

## Proposed fix

1. Add `d.HasChange("memory")` and `d.HasChange("cpu")` gating in Update — only include the field in the request if it actually changed.
2. Introduce a new attribute `allow_vm_restart` (Bool, Optional, Default: false) on `cloudtemple_compute_virtual_machine`. Documented as provider-side only — does NOT go in the API payload.
3. If `memChanged && !hot_add_enabled && !allow_vm_restart`, return `diag.Errorf` with an actionable message BEFORE calling the API: "changing memory requires a power cycle ; set allow_vm_restart=true to allow the provider to power-off, PATCH, power-on automatically, or change memory_hot_add_enabled to true".
4. If `allow_vm_restart=true`, orchestrate the power-off → PATCH → power-on sequence locally (the code already does this for `customize`, see lines 1347-1397).
5. The same gating applies to `cpu` and `num_cores_per_socket`.

## Mutation-proof tests

- Verify that an Update with no change to memory does NOT include `Ram` in the payload (mock client).
- Verify that an Update with memory change + hot-add OFF + `allow_vm_restart=false` fails fast with the actionable diag (no API call).
- Verify that `allow_vm_restart=true` orchestrates the power cycle and the state converges.

## Refs

- Parent investigation: #392
- `.clinerules/WORKFLOW_ENGINEERING.md` §5 state-safety
- Epic: #387 (v1.9.0 hardening)
- Related: #339 (provider-side resilience)

Alexandre-Cann:
Closes #238
Closes #374
Closes #375
Closes #379
Closes #382
Closes #384
Closes #386
Closes #387
Closes #394
Closes #395
Closes #396
Closes #397

## Release v1.9.0 — VPC integration + provider hardening

VPC provider surface (datasources + resources), the 403/404 not-found contract alignment, and a batch of Compute state-safety hardening.

### ⚠️ Upgrade note
Requires **Compute API >= v1.144.0** in production: Compute reads now treat a 404 as absent and surface a genuine 403 as access-denied (#384). The 404-on-absent contract was verified live against production on all 29 Compute endpoints before this release.

### Highlights
- **VPC**: datasources `cloudtemple_vpc_{vpc,private_network,static_ip,floating_ip}` (+ plural), resources `cloudtemple_vpc_static_ip`, `cloudtemple_vpc_floating_ip`, `cloudtemple_vpc_floating_ip_binding`; `ip_address` on VMware & OpenIaaS network adapters for VPC attachment (#238, #374, #375, #379).
- **403/404 contract**: Backup/Marketplace/VPC and Compute (VMware + OpenIaaS) reads now surface a real 403 as access-denied instead of folding it to not-found (#384).
- **State-safety hardening**: no silent memory/cpu/cores shrink (#395), os_disk shrink guard (#394), opt-in power-cycle on running-VM resize via `allow_vm_restart` (#397), OpenIaaS sizing re-assert + read-back (#396).
- **Crash fixes**: datasource nil-deref guard (#382), VMware VM op nil-deref guard (#386).

Out of scope (stay open): #391 (platform-side backup-policies hang, provider mitigation only), #393 (test-complacency follow-up).


Alexandre-Cann:
## Release validation (v1.9.0) — for the reviewer

**RC gate — `ct-validate probe_absence` run live against PRODUCTION (`shiva.cloud-temple.com`):**
- All **29 Compute endpoints** (17 `compute.vmware.*` + 12 `compute.openiaas.*`) return **404-on-absent**. **Zero 403** anywhere (Backup, Marketplace, VPC also 404). → the #384 403→404 flip is safe on the live contract.
- Prod Compute API = **v1.144.1** (≥ the required v1.144.0 in the UPGRADE NOTES).
- `object_storage.{bucket,storage_account}` returned **HTTP 500** — a pre-existing platform anomaly, **out of scope**: those reads still use `requireNotFoundOrOK(resp, 403)` (not part of the #384 flip), so v1.9.0 changes nothing for object storage. Candidate for platform feedback, non-blocking.

**Diff nature:** the merge commit's net change vs the previous `rc/v1.9.0` tip is **`CHANGELOG.md` only** (date the 1.9.0 section, move the `latest` badge, pull the v1.7.2/#320 hotfix). No code delta. Local `go build ./...` green; CI green (Build, Unit tests & vet, generate).

**Scope:** closes the 12 *Awaiting release* issues listed in the description. **#391** (platform-side backup-policies hang — provider mitigation only) and **#393** (test-complacency follow-up) intentionally stay **open** (board status: Todo).

KChapron (approved):

PR #503, opened 2026-07-21
fix(#488): stop capturing the cloud-init config drive as a managed os_disk

pbesret:
Refs #488

## Problem

A `cloudtemple_compute_iaas_opensource_virtual_machine` deployed with `cloud_init` captured
the platform's cloud-init config drive ("XO CloudConfigDrive") as a managed `os_disk` at
create time. Every subsequent plan then proposed to remove that block — a permanent,
non-converging drift.

## Root cause (live-proven)

The #255/#273 guard excludes the drive on **exact name AND read-only VBD**. Live evidence
(raw API capture, DEV tenant, 2026-07-21) shows the platform attaches the drive with a
**read-write** VBD on the deploy path (`"readOnly": false`), and never auto-destroys it, so
the guard never matches. The drive is listed before the deploy activity completes (the
platform configures cloud-init before completion on both the marketplace and template
paths), so the capture is systematic, not a race.

Reproduced identically with released **v1.8.0** and HEAD: this is a platform-behavior gap,
not a v1.9.0 provider regression (the create path and the guard are byte-identical between
those versions).

## Fix (validated by a 4-round adversarial plan review)

**Create** (no pre-existing managed entry can be lost there):
- preflight: a create is rejected upfront when the source image (template or marketplace
  item) itself carries the reserved disk name while cloud-init is provisioned —
  evidence-only, absent/empty metadata passes;
- the flatten skips the platform drive on: exact name + VBD of this VM + (read-only VBD OR
  the resource's own cloud-init provisioning, with the exact `buildOpenIaasCloudInit`
  payload semantics);
- ambiguity guard: two or more discriminator matches fail the create closed, naming the ids.

**Read** (state safety):
- the automatic drop keeps the strict #255 evidence (name + read-only VBD) — a state entry
  is **never** dropped on the weaker evidence;
- a captured RW drive (state and live both carry the reserved name, cloud-init provisioned)
  is kept and a warning explains the remediation: applying the removal diff cleans the
  Terraform state only — the provider never detaches, renames or resizes the disk
  (`handleUpdateOSDevices` acts only on disks still present in the desired list).

**Schema**: "XO CloudConfigDrive" is rejected as a user-declared `os_disk.name`
(config-only validation; polluted states keep planning and self-heal).

## Evidence

- Unit: discriminator matrix (both evidences, VBD fail-safe, both listing orders), count
  guard, preflight rules, provisioning-flag semantics through the real schema,
  resource-level reserved-name validation, JSON decode contract-pin for
  `virtualMachines[].readOnly` (the client structs carry no JSON tags).
- Mutation proof: reverting the discriminator to the pre-fix `return vbd.ReadOnly` turns
  exactly the new tests RED (recorded in the review thread).
- Live E2E on the DEV tenant (az05, 2026-07-21), all PASSED:
  - (a) create with cloud_init using this branch: the state holds exactly ONE `os_disk`
    (the system disk) while the raw API lists TWO disks attached (system + the RW
    "XO CloudConfigDrive"); the immediate `terraform plan -detailed-exitcode` exits 0
    ("No changes") — the issue's idempotence expectation is restored;
  - (b) remediation: a state carrying the captured drive (real platform values, the exact
    v1.8.0/v1.9.0/v1.10.0 pollution shape) plans with the guidance warning
    ("Captured cloud-init config drive present in os_disk") plus the removal diff; one
    apply purges the entry (state back to one block), the next plan is empty, and the
    platform still reports the drive ATTACHED afterwards — nothing was detached;
  - baseline: the same deploy through the RELEASED v1.8.0 provider reproduces the capture
    (two os_disk blocks) when the create-time race is lost, confirming the pre-fix drift
    and its non-deterministic "sometimes v1.8.0 looks fine" behavior.

## Residual (documented)

The create-time skip relies on the platform completing the deploy activity only after the
config drive exists (verified in the worker source for both paths) when source-disk
metadata is absent. A deliberately colliding disk attached out-of-band stays simply
unmanaged, as any out-of-band disk (the Read never adopts unknown disks).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ***Warning: Using "Release Candidate" versions (-rc.X) in a **production environment** is **strongly discouraged**, as they may contain unresolved bugs and pose risks to the stability and security of your systems.***
 
+# 1.11.0 (Unreleased)
+
+BUG FIXES :
+
+  * `cloudtemple_compute_iaas_opensource_virtual_machine`: a VM deployed with `cloud_init` no longer captures the platform's cloud-init config drive ("XO CloudConfigDrive", attached read-write by the platform during the deploy) as a managed `os_disk`, which made every subsequent plan propose to remove that block (permanent drift). The create now excludes the drive (and fails closed if the listing is ambiguous), and a create from a source image that itself carries the reserved disk name is rejected upfront. States already polluted by earlier versions keep planning normally: the refresh emits a warning explaining that applying the removal diff only cleans the Terraform state — the provider never detaches, renames or resizes the drive on the platform. The name "XO CloudConfigDrive" is now rejected as a user-declared `os_disk.name` (reserved by the platform).
+
 # 1.10.0 (July 17th, 2026)
 <img id="latest" src="https://badgen.net/badge/channel/latest/yellow" alt="Channel: latest" />
 
--- a/docs/resources/compute_iaas_opensource_virtual_machine.md
+++ b/docs/resources/compute_iaas_opensource_virtual_machine.md
@@ -222,7 +222,7 @@ Order of the elements in the list is the boot order.
 Optional:
 
 - `connected` (Boolean, Deprecated) Whether the disk is connected or not.
-- `name` (String) The name of the operating system disk. (Updating this property implies a disk disconnect-reconnect)
+- `name` (String) The name of the operating system disk. (Updating this property implies a disk disconnect-reconnect) The name "XO CloudConfigDrive" is reserved by the platform for the cloud-init config drive and cannot be used.
 - `size` (Number) The size of the operating system disk in bytes. (Updating this property implies a disk disconnect-reconnect)
 - `storage_repository_id` (String) The storage repository where the operating system disk is located.
 
--- a/internal/client/compute_iaas_opensource_virtual_disk_decode_test.go
+++ b/internal/client/compute_iaas_opensource_virtual_disk_decode_test.go
@@ -1,0 +1,73 @@
+package client
+
+import (
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+)
+
+// TestOpenIaaSVirtualDiskDecodeContract pins the deserialization of the
+// virtual-disk payload against the LIVE response shape (captured 2026-07-21
+// on the DEV tenant while reproducing #488). The structs carry no JSON tags:
+// decoding relies on Go's case-insensitive field matching, so a silent API
+// field rename (readOnly -> read_only, …) would zero the field without any
+// compile-time signal. IsPlatformManagedDisk keys the platform-drive
+// exclusion on VirtualMachines[].ReadOnly — this test is the tripwire.
+func TestOpenIaaSVirtualDiskDecodeContract(t *testing.T) {
+	const payload = `{
+		"id": "1f5c8276-99b2-4e16-9634-8f1ad51e4993",
+		"internalId": "3fe6abe0-028a-463f-91d9-3e6781778608",
+		"name": "XO CloudConfigDrive",
+		"description": "",
+		"size": 10485760,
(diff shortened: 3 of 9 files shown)


pbesret (commented):
**Independent model review — verdict recorded verbatim** (reviewer: Codex gpt-5.5, reasoning effort high, read-only sandbox; pilot: Claude. Full cycle: plan reviewed adversarially over 4 rounds — 3 NO-GO hardening the design, then GO — plus a pre-commit GO and this final pass on the committed diff with the complete evidence.)

---

No blocking findings.

Regression risk I can still construct: only the intentional one. Existing configs that explicitly declare `os_disk.name = "XO CloudConfigDrive"` will now fail validation. That is documented and consistent with the new reserved-name contract. For state safety, I do not see an unhandled destructive path:

- Clean states stay clean; read only walks existing `os_disk` state entries.
- Polluted RW #488 states are kept with warning, then the removal apply is state-only.
- Strict #255 auto-drop still requires exact name + read-only VBD on this VM.
- Renamed/out-of-band disks stay managed because state-name provenance blocks the warning classifier.
- Source-collision creates are rejected pre-deploy when source metadata proves the reserved name, and ambiguous post-deploy listings fail closed.

Release train scope looks complete: code, unit coverage, mutation pinning, live E2E evidence, CI, changelog under `1.11.0 (Unreleased)`, regenerated resource docs, schema snapshot update, and coverage floor note are all present. I do not see an RC validation tripwire left from this diff.

I did not rerun tests locally because this workspace is read-only, but the supplied CI/live evidence is sufficient for merge readiness.

VERDICT: GO

---

Mandatory human review and the human RC gate still apply before merge, per `WORKFLOW_GIT_EPIC.md`.

Alexandre-Cann (commented):
**Independent review — GO (non-blocking)**

Scope: full PR diff (`fc0cef2`, +537/−59, 9 files), surrounding code at head, call-site sweep, `.clinerules` conformance.

**Checks performed**

- State-safety: the automatic drop still requires the strict #255 evidence (exact name + read-only VBD); the weaker RW+cloud-init evidence only yields keep+warning — never a drop. Fail-safe on no-VBD / other-VM / empty vmID. ✔
- Create path: the skip is confined to the flatten where no pre-existing managed entry can be lost; the ambiguity guard fails closed naming the ids; the preflight is evidence-only (absent/empty source metadata passes). ✔
- Call-site sweep: `FlattenOpenIaaSOSDisksData` / `IsPlatformManagedDisk` have no other non-test callers; both call sites updated. ✔
- Tests: discriminator matrix incl. both listing orders, VBD fail-safe, empty vmID, provisioning-flag semantics through the real schema, real-schema validation pin, JSON decode contract-pin on `readOnly`. Non-vacuous; mutation proof recorded. ✔
- Process: RC-flow conventions respected (base `rc/v1.11.0`, `Refs #488`), CHANGELOG entry present, coverage floor raised 70.0→70.9 (never lowered), schema snapshot updated. ✔

**Findings (non-blocking)**

1. MINOR — the client structs carry no JSON tags; the decode contract test is a good tripwire, but explicit `json:"readOnly"` tags would remove the risk class rather than detect it. Suggest a follow-up issue.
2. MINOR — `classifyOSDiskOnRead`: a captured drive whose state entry has an empty/diverging `name` falls back to plain keep with no guidance warning. Non-destructive and self-correcting (the next flatten rewrites the live name, so the warning appears one refresh later). No change requested.
3. NOTE — the create-time skip relies on the platform attaching the drive before deploy-activity completion (verified in worker source, documented as residual). Recommend keeping E2E (a) in the RC validation as a permanent regression probe.

**Limitations**: reviewed from a sandbox without a Go toolchain — the test suite was not executed by this reviewer; imports and signatures verified manually, execution evidence delegated to CI.

Verdict: **GO** — mergeable once CI is green, per the human-gate rules.

Issue #506, opened 2026-07-23
public_cloud_vm_instance_families/instance_family: expose SKU pricing (skus)


pbesret:
## Context

The VM Instances API now returns a `skus` array on each instance family (list and single-item endpoints), carrying the priced SKUs of the family resources (vCPU and RAM):

```json
"skus": [
  { "name": "csp:fr1:vminstance:gp:vcpu:v1", "price": 23.42, "unit": "vcpu", "description": "…", "descriptionEn": "…" },
  { "name": "csp:fr1:vminstance:gp:ram:v1",  "price": 2.56,  "unit": "gio",  "description": "…", "descriptionEn": "…" }
]
```

## What to do

Expose it on `cloudtemple_public_cloud_vm_instance_families` and `cloudtemple_public_cloud_vm_instance_family`: new read-only `skus` (List of Object: `name` String, `price` Number, `unit` String, `description` String, `description_en` String) in the nested instance family schema — client model, datasource schemas, regenerated docs, acceptance tests. Required role unchanged (`public_cloud_vm_instances_read`).

## Dependency

Requires the next VM Instances API release (contract visible in the API swagger, `InstanceFamily.skus` / `Sku`).

PR #508, opened 2026-07-23
feat(#506): expose instance family SKU pricing (skus) on the two datasources

pbesret:
Refs #506

## What

Exposes the priced billing SKUs (vCPU and RAM) returned by the VM Instances API
as a new **read-only** `skus` attribute on the two existing instance-family data
sources:

- `cloudtemple_public_cloud_vm_instance_family` (single)
- `cloudtemple_public_cloud_vm_instance_families` (list)

`skus` is a list of objects: `name`, `price`, `unit`, `description`,
`description_en`. The authorization role is unchanged
(`public_cloud_vm_instances_read`). No resource is touched.

## How

- **Client**: new `PublicCloudVMSku` type (explicit `json` tags pinning the
  camelCase API contract, incl. `descriptionEn`) and a `Skus` field on
  `PublicCloudVMInstanceFamily`.
- **Flatten**: emits `skus` **sorted by `name`** (`sort.SliceStable`) so a
  read-only, indexable datasource list has a stable order regardless of API
  order, and returns a non-nil empty slice when a family has no SKUs.
- **Schema**: a shared `publicCloudVMSkusSchema()` helper feeds both data
  sources so they cannot drift apart; `price` is `TypeFloat`.
- Schema snapshot and docs regenerated; examples updated with a `skus` output.

## Why it is safe (additive, read-only)

The field is `Computed` on data sources only — strictly additive. No
`SchemaVersion` bump, no `StateUpgrader`. Existing configurations and states are
unaffected (no one references `skus` yet); a family with no SKUs renders as an
empty list.

## Live API contract (verified)

A read-only live probe against the dev API confirmed both endpoints already
return `skus` with the exact camelCase shape used here. Prices are JSON numbers
(decimal e.g. `23.42` and integer e.g. `2`), decoded as `float64`. The typed
client decodes it live on both `List` (all families) and `Read`.

## Tests

- **Client decode** test pins the exact contract: decimal and integer prices ->
  `float64`, and `descriptionEn` -> `DescriptionEn`.
- **Flatten-level** test proves `skus` is actually emitted and sorted by name —
  a gap the structural walker cannot catch (it only proves emitted shape *fits*
  the schema, not that the key is emitted). Mutation-proven: removing the
  emission, or reversing the sort, turns it RED; the correct code is GREEN.
- **Acceptance** tests assert `skus.#`, `skus.0.name`, `skus.0.price` (and
  `unit`) on both data sources.
- `go build`, `gofmt`, `go vet ./internal/...` and the full offline unit suite
  (client + helpers + provider, incl. the flatten/schema walker, the zero-input
  invariants and the schema snapshot) are green.

## Review

Independent adversarial review (Codex): committed-diff verdict **GO**, no
blocking finding (the two nits — `SliceStable`, and a typed `Read` in the probe —
were applied).

## Note on acceptance execution

The acceptance package's `TestMain` performs a destructive cleanup on the target
environment at start-up (it deletes VMs named `test-terraform*` / `hello-world`
or tagged `created_by=Terraform`). The `skus` acceptance assertions were added
but their live execution is gated on a dedicated environment; live coverage for
this PR comes from the read-only probe.

Awaiting independent reviewer.


pbesret:
Independent adversarial review (Codex, read-only) on the committed diff: **VERDICT GO**, no blocking finding. The two non-blocking points were applied: the flatten sort uses `sort.SliceStable`, and the read-only live probe now also exercises the typed `Read`.

Live evidence (read-only probe against the dev VM Instances API): the typed client decodes `skus` on both `List` (all families) and `Read`; prices decode as `float64` for decimal and integer JSON numbers; `descriptionEn` maps to `DescriptionEn`.

Offline gates green: client decode test (mutation-proven), flatten-level emission+sort test (closes the walker gap; RED under mutation), flatten/schema walker, zero-input invariants, schema snapshot, `go vet`.

Pilot and reviewer act through the same GitHub account, so per the RC review discipline this needs an independent human validation before the RC merge.

Verdict: Awaiting independent reviewer

pbesret:
CI is green on this PR (Build, Unit tests & vet, generate).

Decision on acceptance execution: the live acceptance run is **deferred to the RC→main step** (dedicated environment, real `SHIVA_TEST_*` credentials), because the acceptance package's `TestMain` performs a destructive VM cleanup on the target environment and the required test credentials are not available in the current context. The `skus` acceptance assertions are already in place on both data source tests. Coverage backing this RC merge: the read-only live probe (typed client decodes `skus` on List and Read against the dev API) plus the offline gates (client decode test — mutation-proven, flatten-level emission/sort test, flatten↔schema walker, zero-input invariants, schema snapshot).

No code change requested by this note.

Alexandre-Cann (approved):

Issue #507, opened 2026-07-23
public_cloud_vm_storage_types/storage_type: expose SKU pricing (sku)


pbesret:
## Context

The VM Instances API now returns a `sku` object on each storage type, carrying the priced SKU of the storage resource:

```json
"sku": { "name": "csp:fr1:iaas:storage:bloc:medium:v1", "price": 0.063, "unit": "1 Gio", "description": "…", "descriptionEn": "…" }
```

## What to do

Expose it on `cloudtemple_public_cloud_vm_storage_types` and `cloudtemple_public_cloud_vm_storage_type`: new read-only `sku` object (`name` String, `price` Number, `unit` String, `description` String, `description_en` String) in the nested storage type schema — client model, datasource schemas, regenerated docs, acceptance tests. Required role unchanged.

## Dependency

Requires the next VM Instances API release (contract visible in the API swagger, `StorageType.sku` / `Sku`).

PR #509, opened 2026-07-23
feat(#507): expose the sku pricing object on the storage type datasources

pbesret:
Refs #507

## What & why

The VM Instances API now returns a `sku` object on each storage type, carrying the priced SKU of the storage resource. This exposes it, **read-only**, on both storage-type data sources:

- `cloudtemple_public_cloud_vm_storage_type` (at the root)
- `cloudtemple_public_cloud_vm_storage_types` (inside each `storage_types` element)

New nested read-only attributes: `sku.name` (String), `sku.price` (Number), `sku.unit` (String), `sku.description` (String), `sku.description_en` (String). **Required role unchanged.** Additive, backward-compatible.

## Changes

- **client** (`internal/client/public_cloud_vm_storage_type.go`): new `PublicCloudVMSku` struct with **explicit json tags** (to lock the exact API spelling of the billing fields, notably the FR/EN description pair) and a **pointer** `Sku` field, so an omitted or `null` sku decodes to `nil` rather than a phantom zero-priced object.
- **helper** (`helper_public_cloud_vm_storage_type.go`): a nil `Sku` flattens to an empty list, a present one to a single-element list — the nested-single-object idiom already used by `publicCloudVMInstanceRefSchema`.
- **schema**: a shared `publicCloudVMStorageTypeSkuSchema()` block (`TypeList`, computed) used by both data sources so the nested shape cannot drift between them. `price` is the provider's **first `schema.TypeFloat`**.
- **generated**: schema snapshot regenerated (**additive, +64 / −0**, only the two `sku` blocks); docs regenerated via `go generate`.
- **CHANGELOG**: `1.11.0 (Unreleased)` → `ENHANCEMENTS`.

## Test evidence (non-complacent)

- **Mutation proof (RED-without-fix):** reverting the helper to its pre-#507 body (no `sku` key) makes `TestFlattenPublicCloudVMStorageTypeSku` fail with *"flatten emits no \"sku\" key; the datasource cannot expose the SKU"*; restoring the fix turns it green.
- **Client decode:** `TestPublicCloudVMStorageTypeList` (sku present, with a float tolerance) + `TestPublicCloudVMStorageTypeListSkuAbsent` (sku **omitted** and explicit **null** → nil `Sku`).
- **Live evidence:** a read-only probe against the dev API confirmed `sku` is returned on 2/2 storage types and decoded by the production client, with the exact wire shape (`price` as a JSON number, `descriptionEn` spelling). The throwaway probe is **not** committed.
- **Auto-gates green:** `TestFlattenZeroInputInvariants` (nil `Sku` → non-nil empty list, no `knownNilSliceGaps` entry needed) and the flatten↔schema walker `TestDatasourceFlattenOutputsFitTheirSchemas` (first `TypeFloat`, and the nested list-under-list for the plural).
- **Acceptance:** both data-source acceptance tests assert `sku.0.name` / `sku.0.price` presence (`TestCheckResourceAttrSet`) — not the volatile live price.

## Design notes

- `sku` is modelled as a computed `TypeList` **without `MaxItems: 1`**, matching the existing computed nested-single-object convention in this domain (`publicCloudVMInstanceRefSchema`: `availability_zone`, `backup_policy`, …). Consistency over decorative precision.
- The schema-snapshot change is purely additive: no reordering, no deletions, no state-shape change → no `StateUpgrader`, `TF prod breaker = NO`.

## Residual risks (accepted, per the confirmed contract)

- The plural acceptance test assumes `storage_types.0` carries a `sku`. This holds for an API that returns `sku` across the catalogue (confirmed live). Against an older/mixed API that omits `sku`, the provider still behaves correctly (empty block), but that acceptance assertion would fail — expected, since #507 depends on the API release that returns `sku`.
- A present `sku` with a `null` scalar (e.g. `price: null`) would decode to a zero value. The contract (issue + live) has `price` as a JSON number, so this is not exercised.

## Review trail

- Adversarial **PLAN** review (Codex, read-only): `VERDICT: GO`, 0 blocking.
- Adversarial **pre-commit** review of this exact diff (Codex, read-only): `VERDICT: GO`, 0 blocking, 2 minor + 1 nit (addressed/accepted above).

## RC flow

Targets `rc/v1.11.0` with `Refs #507` (not `Closes` — the closing keyword is reserved for the RC → `main` PR). No tag/release from the RC branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # 1.11.0 (Unreleased)
 
+ENHANCEMENTS :
+
+  * The `cloudtemple_public_cloud_vm_storage_type` and `cloudtemple_public_cloud_vm_storage_types` data sources now expose the priced `sku` object the VM Instances API returns on each storage type: `sku.name`, `sku.price`, `sku.unit`, `sku.description` and `sku.description_en` (all read-only). The `sku` block is empty when the API returns no SKU for a given storage type. This requires a VM Instances API version that returns `StorageType.sku`; against an older API the block is simply empty (#507).
+
 BUG FIXES :
 
   * `cloudtemple_object_storage_bucket`: the resource now refreshes `access_type` from the API, so a change made out-of-band in the Cloud Temple console (for example `custom` -> `private`) is detected as drift instead of staying invisible to `terraform plan` (#490). Previously the read path never wrote `access_type` back to the state, so console-side changes were silently ignored. Existing states whose console value has already diverged will show a one-time reconciling diff on the next plan — the expected behaviour, letting `terraform apply` realign the intended value. The `cloudtemple_object_storage_bucket` and `cloudtemple_object_storage_buckets` data sources now also expose `access_type`.
--- a/docs/data-sources/public_cloud_vm_storage_type.md
+++ b/docs/data-sources/public_cloud_vm_storage_type.md
@@ -46,5 +46,17 @@ output "storage_type" {
 - `is_available` (Boolean) Whether the storage type is currently available.
 - `max_size_gb` (Number) The maximum disk size (GB) allowed for this storage type.
 - `min_size_gb` (Number) The minimum disk size (GB) allowed for this storage type.
+- `sku` (List of Object) The priced SKU of the storage resource. Empty when the API returns no SKU for this storage type. (see [below for nested schema](#nestedatt--sku))
+
+<a id="nestedatt--sku"></a>
+### Nested Schema for `sku`
+
+Read-Only:
+
+- `description` (String)
+- `description_en` (String)
+- `name` (String)
+- `price` (Number)
+- `unit` (String)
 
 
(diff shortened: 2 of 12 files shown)


pbesret:
## Adversarial review trail (Codex, read-only)

**PLAN review** (before coding): `VERDICT: GO`, 0 blocking. Confirmed the pointer `*Sku` design (avoids a phantom zero-priced object), the nil→empty-list flatten, `TypeFloat` for `price`, and no hidden coupling (the helper is used only by these two data sources).

**Pre-commit review** (this exact diff): `VERDICT: GO`, 0 blocking, 2 minor + 1 nit — all validated against the code and resolved/accepted:
- *minor* — the plural acceptance test assumes `storage_types.0` carries a `sku`: holds for the API release that returns `sku` across the catalogue (verified live); documented as a residual.
- *minor* — a present `sku` with a `null` scalar (`price: null`) would decode to zero: the contract (issue + live) has `price` as a JSON number, so it is not exercised.
- *nit* — no `MaxItems: 1` on the `sku` block: deliberate, to match the existing computed nested-single-object convention (`publicCloudVMInstanceRefSchema`).

Pilot and reviewer act through the same GitHub account (model-role separation). `Awaiting independent reviewer` — an independent human validation is required before merge; the merge itself is held for the human RC gate.

Alexandre-Cann (approved):

Issue #513, opened 2026-07-30
public_cloud_vm_storage_types/storage_type: paired availabilityZoneId + instanceFamilyId list filters


pbesret:
## Context

The VM Instances API update adds two list filters to `GET /vm_instances/v1/storage_types`: `availabilityZoneId` and `instanceFamilyId`. They are optional, but **paired**: a request carrying only one of the two is rejected by the API (opaque HTTP 500, live-verified), while a request with none or both answers 200.

## What to do

Expose the pair on `cloudtemple_public_cloud_vm_storage_types` and `cloudtemple_public_cloud_vm_storage_type`: new optional `availability_zone_id` and `instance_family_id` arguments (UUID-validated), enforced as a pair at plan time (`RequiredWith` both ways) so a lone filter never reaches the API; on the singular datasource the pair narrows the catalogue in which `id`/`name` is resolved. Client model filter + fail-closed guard, datasource schemas, regenerated docs, acceptance tests. Required role unchanged.

Note: the sibling `/vm_instances/v1/images` endpoint accepts a lone filter (HTTP 200, live-verified), so the pairing constraint is **not** applied to the image data sources.

## Dependency

Requires the VM Instances API version that filters `/vm_instances/v1/storage_types` by `availabilityZoneId` + `instanceFamilyId`; against an older API the filters are silently ignored server-side (unknown query names are not rejected).

Issue #510, opened 2026-07-24
Rename Public Cloud VM template surface to image (breaking)


pbesret:
## Context

The VM Instances API renamed its OS-image catalogue endpoint from `GET /vm_instances/v1/templates` to `GET /vm_instances/v1/images`, and the VM contract from `templateId` / `template{id,name}` to `imageId` / `image{id,name}`. The provider must follow this rename on its `public_cloud_vm_*` surface.

## Scope

Rename `template` to `image` across the Public Cloud VM Instances surface (breaking, no backward-compatible alias):

- data sources `cloudtemple_public_cloud_vm_template` / `..._templates` → `cloudtemple_public_cloud_vm_image` / `..._images`;
- their `template_type` attribute → `image_type`, and the `templates` list attribute → `images`;
- resource `cloudtemple_public_cloud_vm_instance`: argument `template_id` → `image_id`, computed `template_name` → `image_name`;
- data sources `cloudtemple_public_cloud_vm_instance` / `..._instances`: computed `template` block → `image`;
- client SDK: endpoint `/vm_instances/v1/templates` → `/images`, create body `templateId` → `imageId`, VM read `template{}` → `image{}`.

A `SchemaVersion` v0→v1 `StateUpgrader` renames the stored `template_id` / `template_name` keys to `image_id` / `image_name`, so a VM already in state is **not** destroyed or recreated. Users must still rename the argument in their HCL.

## Out of scope

`compute_iaas_opensource_template*` (a separate product) is untouched.

## Release gating

`GET /vm_instances/v1/images` is deployed on DEV only today; the release to `main` is gated on the PROD broker switch. This change lands on the `rc/v1.11.0` train.

PR #515, opened 2026-07-31
chore(release): date the 1.11.0 CHANGELOG section and add UPGRADE NOTES

pbesret:
Release preparation for **v1.11.0**. CHANGELOG only — no code change.

### What this does

- Dates the `1.11.0` section (July 31st, 2026) and moves the `latest` badge from 1.10.0 to 1.11.0.
- Adds an `UPGRADE NOTES` section gathering the four things a user must know before upgrading. Those prerequisites already existed, but were buried at the end of individual entries where a reader planning an upgrade is unlikely to find them.
- Tags the cloud-init config drive bug-fix entry with its issue number (#488), which was the only entry missing one.

### The #510 note, and the evidence behind it

PR #512 gated this release on a "PROD broker switch", because `/vm_instances/v1/images` was DEV-only at the time. A read-only probe (4 GETs, no write) of the production endpoint on July 31st, 2026 returned:

```
GET /vm_instances/v1/images                -> 403 "Forbidden."
GET /vm_instances/v1/storage_types         -> 403 "Forbidden."
GET /vm_instances/v1/zzz_nonexistent_route -> 404 "Cannot GET /v1/zzz_nonexistent_route"
GET /vm_instances/v1/templates             -> 404 "Cannot GET /v1/templates"
```

The probing tenant is not entitled on the Public Cloud VM catalogue, which is why the catalogue paths answer 403. The routing discriminant is what makes the result usable: a path that does not exist answers 404 from the downstream service, while `/images` answers 403. `/images` is therefore served on that endpoint — the gate is cleared.

The same evidence shows `/templates` answering the 404 of a non-existent path. On that endpoint the pre-1.11.0 image surface no longer resolves, so this release repairs an existing breakage rather than introducing a new risk.

The CHANGELOG note is deliberately scoped to what was measured — one endpoint, one date, one method — and gives a reader on a different endpoint the criterion to check their own. It does not generalise to every deployment.

### Review

Independent adversarial review (Codex, read-only), two rounds. Round 1 returned NO-GO on exactly that point: the first draft claimed "this upgrade is required, not optional" and stated universal production removal from a single-tenant probe, which could push a user still served by `/templates` into a breaking rename. Both blocking findings were accepted and fixed; two minor findings were accepted (the `(#488)` tag, and separating input validation from server-side filtering in the #513 note); one minor was declined with justification (British spelling is this file's existing convention, not a drift). Round 2: GO, no findings.

### Checks

- `go vet ./...` clean; no code touched, so the test surface is unchanged.
- Exactly one `latest` badge in the file, on 1.11.0.
- No `Unreleased` marker left.


pbesret:
CI green on this branch: Build, Unit tests & vet, and generate all pass.

Ready to merge into `rc/v1.11.0`. Merging is the human gate — once it lands, the RC tip moves and the `rc/v1.11.0` -> `main` release PR can be opened on the final SHA, with the `RC-VALIDATION: OK` marker published against that SHA (a marker published before this merge would be invalidated by it).

Alexandre-Cann (approved):

Issue #473, opened 2026-07-02
E9-4 resource VM: inline os_network_adapter must refuse VPC networks (phase 1)


pbesret:
**Epic:** Refs #409  ·  **Lot:** S8 Public Cloud VM  ·  **Plan:** `E9-4`  ·  **Milestone:** v1.10.0

## Objective
Phase 1 of the VPC integration: the VM resource's inline `os_network_adapter` block must refuse VPC networks with an actionable error pointing to the standalone `cloudtemple_public_cloud_vm_network_adapter` resource.

## Acceptance criteria
- Create preflight resolves each inline `network_id` and refuses when the network is VPC (fail-closed when the network cannot be read); unit tests; docs updated. Depends on E9-2.

Issue #376, opened 2026-06-28
feat(vpc): create a VM with its inline os_network_adapter on a VPC network (VM Instances, OpenIaaS, VMware)


pbesret:
Refs #238. Supersedes the original OpenIaaS-only scope of this issue: the same capability is now tracked here for **all three VM surfaces**.

## Goal

Create a virtual machine whose **inline `os_network_adapter` block** is attached to a **VPC private network**, in a single `terraform apply` — optionally with a **chosen static IP** — on:

- `cloudtemple_public_cloud_vm_instance` (VM Instances / public cloud)
- `cloudtemple_compute_iaas_opensource_virtual_machine` (OpenIaaS)
- `cloudtemple_compute_virtual_machine` (VMware / vCenter)

Until now this required a two-step path on every surface: create the VM on a Private Backbone network, then attach or re-address the adapter with the standalone `cloudtemple_*_network_adapter` resource. On the VM Instances resource a VPC-only machine was not expressible at all, because the inline block is `Required` with `MinItems: 1`.

## What actually blocked it, per surface

The three surfaces were blocked by **different** things, which is why this needs one issue rather than three unrelated ones:

| Surface | Blocker |
|---|---|
| VM Instances | A deliberate provider-side refusal (the "VPC phase 1" rule of #473). The client already sent `networkInterfaces[].ipAddress`; nothing on the API side was missing. |
| OpenIaaS | A genuine API gap: the VM-create body's `networkAdapters[]` was documented as accepting only `{networkId, mac}`. |
| VMware | Neither a refusal nor an API gap — simply no `ip_address` argument on the inline block, although the adapter PATCH the resource already performs accepts one. |

## Status of the API gap (the reason this was Blocked)

**The OpenIaaS API gap is closed.** Verified against the live API rather than inferred from the broker schema: a create carrying `networkAdapters[].ipAddress` on a VPC-backed network registers exactly that address for the adapter's MAC. Omitting it lets the platform auto-assign one, and deleting the VM reclaims the registration.

This issue can therefore leave `Blocked`.

## Acceptance criteria

- [ ] On each of the three resources, an inline `os_network_adapter` block can target a VPC-backed network in one apply.
- [ ] An optional `ip_address` on that block registers the chosen VPC static IP; omitting it lets the platform auto-assign.
- [ ] `ip_address` is refused when the target network is not VPC-backed, since the platform silently discards it there — a chosen address must never be dead configuration the user believes took effect.
[...]

PR #523, opened 2026-08-20
feat(#376): create a VM with its inline os_network_adapter on a VPC network (VM Instances, OpenIaaS, VMware)

pbesret:
Refs #376

Creates a virtual machine whose **inline `os_network_adapter` block** sits on a **VPC private network**, in a single `terraform apply`, with an optional **chosen static IP** — on all three VM surfaces.

## What blocked each surface, and what changed

| Surface | Blocker | Change |
|---|---|---|
| `cloudtemple_public_cloud_vm_instance` | A deliberate provider refusal (the "VPC phase 1" rule of #473) | Refusal removed. **No client and no schema change**: the client already sent `networkInterfaces[].ipAddress`. A VPC-only machine becomes expressible for the first time. |
| `cloudtemple_compute_iaas_opensource_virtual_machine` | A real API gap — the VM-create `networkAdapters[]` took only `{networkId, mac}` | Gap is closed (verified live). `ip_address` added to the inline block and `IPAddress` to `client.OSNetworkAdapter`. |
| `cloudtemple_compute_virtual_machine` | Neither: simply no `ip_address` argument | `ip_address` added. **No client change** — the address rides the adapter PATCH the resource already performs. |

The decision logic lives once, in `internal/provider/vpc_inline_adapter_ip.go`, and is shared by the two Compute resources; nothing is forked between surfaces.

## Guards, and why each one exists

Each is there because of something the platform actually does, measured on DEV:

- **`ip_address` requires a VPC-backed network.** On a plain network the platform *silently discards* it — no registration, no error. Without this guard a chosen address is dead configuration the user believes took effect.
- **`ip_address` cannot be used when another inline block targets the same network.** The platform assigns an explicit address per (virtual machine, network) pair and refuses otherwise. Left to the platform, that refusal arrives from the create activity **and leaves an untracked virtual machine behind** — observed, and cleaned up by hand. The preflight removes the provider-triggered cause.
- **`ip_address` is refused on a deployment mode that produces no adapter** — a VMware from-scratch create, an OpenIaaS marketplace deploy — rather than silently ignored.
- **Every refusal precedes any platform change**, and a refused apply records none of the planned values in the state.

## The state-safety property, made structural

`ip_address` is **write-only**: the platform never echoes the registration on the adapter object (it is addressable only by MAC on the VPC plane), so the read carries the previous state value across instead of re-deriving it. Reading it back by MAC would put a `/vpc/v1` call on the **refresh** path of every VM and break `terraform refresh` for a token without `vpc_read`; losing drift detection on this one attribute is the cheaper trade.

That choice has a sharp consequence. terraform-plugin-sdk/v2 persists the **planned** values when a create or update returns an error, and because the read preserves them, a single unrolled error return is enough to record an address that was never applied — after which state equals config, Terraform sees no diff, the function is never called again, and the address is **never applied**. Permanent, and invisible in a plan.

Wrapping each error return was tried and did not hold: review found five separate forgotten spots, each narrower than the last. So the four entry points now use a named diagnostics result and install one guard:

```go
defer rollbackInlineAdapterIPsOnAnyError(d, &diags)
```

It acts on what the function *returns*, so every present and future error path is covered, and all the ad-hoc wrappers are gone — `restoreInlineAdapterIPsOnFailure` has exactly one caller. Rolling back on **any** error, including after the address applied, is deliberate: the reconciliation compares the configuration against the **live** platform rather than the state, so an over-reverted address costs one no-op apply while under-reverting costs correctness permanently.

An AST-backed guard test keeps it installed: for each entry point it requires a named `diag.Diagnostics` result and the `defer` **before the first return**. AST rather than string matching, so a `defer` inside a comment cannot satisfy it.

## Verification

**Live on DEV, through real Terraform.** Every surface has a full create-with-a-chosen-address cycle plus a `PlanOnly` step proving no perpetual diff:

| Test | Result |
|---|---|
| `TestAccResourcePublicCloudVMInstanceVPCInline` | PASS |
| `TestAccResourceIaasOpensourceVirtualMachineVPCInline` | PASS |
| `TestAccResourceVirtualMachineVPCInline` | PASS |
| `...IPOnNonVPCRejected` (all three surfaces) | PASS |
| `...IPOnSharedNetworkRejected` | PASS |
| `...IPAddedOnUpdateRejected` | PASS |
| `...IPRejectedBeforeAnyMutation` (rename + invalid address, then an empty plan) | PASS |
| `...IPReconvergesWithoutAdapterDiff` | PASS |

The create tests assert the **live** registration read by MAC, not merely the state: `os_network_adapter.0.ip_address` is overlaid from the configuration and would match even if nothing had been pushed. Neutering the reconciliation makes the VMware test fail with *"the LIVE registration is 10.0.0.34, want the configured 10.0.0.236"*.

**Non-complacent by construction.** Every guard has a mutation that turns a specific test RED — re-adding the old refusal, dropping a guard, sending `ipAddress` without `networkId`, pushing unconditionally, restoring random map iteration, removing the deferred rollback, commenting it out, moving it after a return, handing it the wrong variable. Each was applied, observed RED, and reverted.

**Orphans.** Post-campaign audit across the compute and IPAM planes: no leftover machine, and every VPC private network back to its pre-campaign static-IP count.

**Gates.** `gofmt` clean, `go vet ./...` clean, `go install` OK, `go generate ./...` leaves no diff, hermetic tests green. Coverage ratchet: `internal/provider` 27.0% (floor 14.3), `internal/client` 40.7% (floor 24.8), helpers 71.7% (floor 70.9). `staticcheck` was not run locally — the installed binary is built with Go 1.23 while the module requires 1.24; CI covers it.

## Schema compatibility

Two strictly additive `ip_address` attributes (Optional, with a validator) inside the existing `os_network_adapter` Elem of the two Compute resources. Nothing removed, renamed or retyped, no `Computed` flag flipped, no `ForceNew` added, `SchemaVersion` unchanged — no existing state shape changes and no `StateUpgrader` is needed. The VM Instances change is Descriptions only, which the golden snapshot deliberately does not capture.

## Upgrade notes for reviewers

- **`ip_address` on a non-VPC network now fails.** It was previously accepted and silently ignored. A configuration carrying one must drop it (the address was never applied) or move the adapter to a VPC network.
- **On `cloudtemple_public_cloud_vm_instance` the inline block stays `ForceNew`.** Now that VPC networks are accepted there, changing an interface's `network_id` or `ip_address` **replaces the virtual machine**. Relocating in place is the standalone adapter resource's job. Removing `ForceNew` would be a breaking schema change and is deliberately out of scope.
- **`vpc_read`** is now documented on the two Compute VM resources, since a configured address reaches that plane.

## Platform findings, routed as feedback rather than tracked here

Five behaviours were measured that belong to the platform, not to the provider, and are being sent through the support feedback endpoint per the repository's rules:

1. `ipAddress` is scoped per (VM, network), not per adapter.
2. An OpenIaaS create activity that fails on that validation **leaves an orphan VM** while reporting no concerned item — so a create failure is not evidence of absence.
3. A static IP whose `source` is `xoa` cannot be deleted through `/vpc/v1`.
4. The VM-image `compatibleTargets` over-promises; the authoritative feasibility signal is the storage-type catalogue for the exact (availability zone, instance family) pair.
5. The tenant carries two OpenIaaS backup policies named `nobackup`, one per machine manager, so resolving that policy by name can pick the wrong one.


pbesret:
Awaiting independent reviewer

## Review trace

Roles were separated per the engineering workflow: the change was piloted by one model and reviewed by an independent adversarial one (Codex, `gpt-5.5`, read-only, non-interactive). Verdicts are recorded as given, not paraphrased.

| Stage | Verdict | Outcome |
|---|---|---|
| PLAN review | NO-GO, then NO-GO | Two rounds. Both blocking sets addressed before any code was written. |
| VM Instances diff | NO-GO → **GO** | Blockers were in the committed live probe, not the feature. |
| OpenIaaS diff | NO-GO ×7 → **GO** | Seven rounds. Every finding was verified against the code before being acted on. |
| VMware diff | NO-GO ×3 → **GO** | Ended in the structural refactor described in the PR body. |
| Final consolidated | **GO** | Two non-blocking suggestions applied. |

## The finding worth reading

The most valuable blocker was not in the feature but in its failure semantics: **a refused apply still wrote the rejected value into the state.** Because `ip_address` is write-only, no refresh corrects it, so state equalled config, no diff remained, and the address was never applied — permanently, and invisibly in a plan.

That was reproduced live before being fixed (`live static IP is 10.0.0.238, want 10.0.0.237`), and the same class then reappeared four more times in narrower places. Patching each return was abandoned in favour of the deferred guard described in the PR body, which is why the fix is structural rather than a sixth patch.

One review round was **partly declined with argument** rather than applied: a proposed rollback of later refusal paths would have been unsound, because those may follow a legitimate mutation and wrapping them could hide a real change. The reviewer accepted that reasoning and pointed at the restructuring instead. Recorded here because a declined finding should be as visible as an applied one.

## What a human reviewer still owns

- The merge decision, and the merge itself.
- Whether `ForceNew` should stay on the VM Instances inline block. Keeping it is deliberate and documented, but it means changing `network_id` or `ip_address` there **replaces the machine**; removing it would be a breaking schema change and was left out of scope.
- Project board placement: this PR and #376 are not moved on the board, since board discipline for this repository is a pending decision rather than something to assume.
- The five platform findings listed in the PR body still need to be sent to the platform team; they are not tracked as provider issues by design.

## Reproducing the live evidence

The acceptance tests are gated on environment variables documented in `.env.test.dist`. Two substrate facts save time on VMware: the from-scratch create produces **no** adapter, so only clone / content-library / marketplace can exercise the feature; and the clone must land on a host cluster whose hosts belong to the distributed switch carrying the VPC portgroup, otherwise the adapter patch fails with `Host <esx> is not a member of VDS <dvs>`. Cloning a machine already on the target VPC network and reusing its own placement avoids both traps.

pbesret:
Superseded by #530.

This pull request was closed as collateral damage when its head branch was renamed away from an agent-generated name (`claude/vm-vpc-network-adapters-695e18` → `feat/376-vpc-inline-adapter-static-ip`). The GitHub rename API does not keep an open pull request attached — that assumption was wrong. No review had been submitted, and the code is unchanged: #530 carries the same three commits plus one more.

Issue #525, opened 2026-09-09
public_cloud_vm: remove the deprecated capacity spellings after the API window closes (2026-12-09)


pbesret:
## Context

The VM Instances API opened a compatibility window for the capacity field rename (`*Gb`/`*Mb` -> `*Gib`/`*Mib`), during which the deprecated spellings are re-emitted as exact aliases. **The window closes on 2026-12-09** (upstream: `applications/shiva/public-cloud/vm-instances!125`).

#524 aligned the provider inside that window: the client accepts both spellings, the read-only attributes were renamed, and the single user-settable attribute was given a deprecation window instead of a hard rename.

This issue tracks what must be removed once the window has closed.

## To remove after 2026-12-09

- [ ] `os_disk.size_gb` on `cloudtemple_public_cloud_vm_instance` — deprecated in v1.12.0, superseded by `os_disk.size_gib`. Removal is a breaking change: it belongs in a major release, with a `SchemaVersion` bump and a state upgrader renaming the stored `size_gb` key to `size_gib` so that no VM is recreated. Note that the state upgrader does not spare users the configuration edit — the CHANGELOG must say so explicitly, as the v1.11.0 `template_id` -> `image_id` entry did.
- [ ] The `ramGb` value accepted by the `order_by` argument of the `cloudtemple_public_cloud_vm_instances` data source.
- [ ] The dual-spelling fallback in the client decode path, once the API no longer serves the old keys against any supported endpoint. Keep the fail-closed behaviour on absence — only the fallback branch goes away.

## Before removing

Confirm against the live API that the deprecated keys are actually gone, rather than relying on the announced date alone: a documented removal date is not evidence of removal. A read-only probe on the affected endpoints is enough.

Refs #524

Issue #524, opened 2026-09-09
public_cloud_vm: align capacity attributes on the API GiB/MiB rename before the 2026-12-09 removal


pbesret:
## Context

The VM Instances API is renaming every capacity field from the `*Gb`/`*Mb` spelling to `*Gib`/`*Mib`, to match the binary units the platform has always used (`1024**3`, `1024`). The rename is a **pure key rename: no value changes**. The old names are re-emitted as exact aliases during a compatibility window that closes on **2026-12-09**, after which they disappear.

Upstream change: `applications/shiva/public-cloud/vm-instances!125` (`tech(#108): align capacity field names on the upstream GiB/MiB rename`).

Renamed keys on the tenant-facing surface:

| Resource | Old key | New key |
| --- | --- | --- |
| Disk | `sizeGb` | `sizeGib` |
| Flavor | `ramGb` | `ramGib` |
| Image | `diskSizesGb` | `diskSizesGib` |
| Instance family | `ramMinGb`, `ramMaxGb` | `ramMinGib`, `ramMaxGib` |
| Quotas | `ramLimitMb`, `storageLimitGb`, `ramUsedMb`, `storageUsedGb` | `ramLimitMib`, `storageLimitGib`, `ramUsedMib`, `storageUsedGib` |
| Storage type | `minSizeGb`, `maxSizeGb` | `minSizeGib`, `maxSizeGib` |
| VM instance | `ramGb`, `disksSizeGb` | `ramGib`, `disksSizeGib` |
| `GET /virtual_machines` | `orderBy=ramGb` | `orderBy=ramGib` |

The write path is **not** affected: the tenant-facing request bodies keep `cpu`, `memory` and `size`; the broker translates them to the upstream spelling.

## Problem

The Public Cloud VM client structs carry no json tags and rely on `encoding/json` case-insensitive matching:

```go
type PublicCloudVMDisk struct {
	SizeGb int // matches "sizeGb", never "sizeGib"
}
```

When the deprecated keys are removed, those fields decode to **0 with no error**. The provider would then write `size_gb = 0`, `ram_gb = 0` and zeroed quotas into the state — a silent failure, not a loud one. On `os_disk.size_gb`, which is `Optional+Computed` and drives the grow-only resize decision, a zero read back from the API pollutes a decision-bearing attribute.

[...]

PR #526, opened 2026-09-09
feat(#524): align the Public Cloud VM capacity surface on the API GiB/MiB rename

pbesret:
Refs #524
Refs #525

Prepares the provider for the VM Instances API capacity rename (`*Gb`/`*Mb` -> `*Gib`/`*Mib`), whose compatibility window closes on **2026-12-09**. Upstream change: `applications/shiva/public-cloud/vm-instances!125`.

## Why this is not just a rename

The rename carries **no value change** — upstream every conversion is binary, so each deprecated name is an exact alias, verified by reading the upstream MR (`data.memory` now goes out as `ram_gib` instead of `ram_gb`, same number).

The danger is elsewhere. The Public Cloud VM client structs carry no json tags and rely on `encoding/json` case-insensitive matching:

```go
type PublicCloudVMDisk struct {
	SizeGb int // matches "sizeGb", never "sizeGib"
}
```

The day the API drops `sizeGb`, that field decodes to **0 with no error**, and the provider writes `size_gb = 0`, zeroed quotas and a zeroed RAM into Terraform state. On `os_disk.size_gb` — `Optional+Computed`, driving the grow-only resize decision — the zero lands in a decision-bearing attribute.

Production still serves **only** the old spelling (the upstream MR targets `dev`), so a hard switch would break the provider today. Hence: accept both.

## What changed

**Client — dual-spelling decode, fail-closed.** Each affected type gains an `UnmarshalJSON` accepting both spellings. An absence under *both* is now an error naming both keys and the offending object, never a zero. The two quota usage counters stay tolerant to absence — the API contract gives them `default: 0` rather than marking them required.

A payload carrying both spellings with **different values** is rejected rather than arbitrated: the deprecated name is contractually an alias, so a divergence means something rewrote the value in transit — a unit conversion applied by a proxy being the dangerous case — and neither value can then be trusted.

**Read-only Terraform surface — renamed (breaking).** The 13 computed attributes move to their binary spelling. Listed in the CHANGELOG under BREAKING CHANGES with an UPGRADE NOTES entry.

**Write surface — deprecation window.** `os_disk.size_gb` is the only user-settable attribute affected, so it is deprecated alongside a new `os_disk.size_gib` instead of renamed. Both drive the same grow-only / VM-must-be-stopped rules; declaring both is rejected at plan time.

**`order_by`** accepts `ramGib` in addition to `ramGb`.

## Two design decisions worth reviewing

**No `ConflictsWith` for the os_disk mutual exclusion.** Both size attributes are `Optional+Computed`, which is exactly why this file already reads `GetRawConfig()` — its own comment says the diff/state view "cannot tell an explicit value from a computed one". `ConflictsWith` resolves against that same unreliable view, so the exclusion lives in `CustomizeDiff` on the raw config, where the rest of the os_disk logic already is.

**No state upgrader for the renamed computed attributes.** Verified in the SDK rather than assumed: `helper/schema`'s `UpgradeResourceState` calls `removeAttributes` **unconditionally** after the upgraders — *"The provider isn't required to clean out removed fields"* — and it recurses into lists, so it reaches inside the `os_disk` block. A computed attribute dropped from the schema is purged from the state and refilled on the next refresh: no plan error, no forced replacement. A state upgrader will be needed when `size_gb` is actually **removed** (#525), not now.

## Schema golden gate — justification for the regenerated snapshot

`internal/provider/testdata/schema_snapshot.json` is regenerated. Per the gate's own policy, the classification:

| Change | Class | Justification |
| --- | --- | --- |
| 13 computed attributes renamed | Breaking / state-shape | Deliberate, documented in BREAKING CHANGES + UPGRADE NOTES. No `StateUpgrader`: the SDK purges attributes absent from the schema (evidence above), and computed values are refilled on refresh. No resource is recreated. |
| `os_disk.size_gib` added (Optional+Computed) | Likely-safe (additive) | New optional attribute; absent from a config, it is simply computed from the API. |
| `os_disk.size_gb` gains `Deprecated` | Non-breaking | Warning text only; the golden does not capture `Deprecated`, so this produces no diff by itself. |

No `Type`, `Elem`, `ForceNew`, `Sensitive`, `Default`, `MaxItems` or `SchemaVersion` change anywhere.

## Known limitation, deliberately accepted

Growing the system disk through `size_gib` leaves the deprecated `size_gb` leaf showing its previous value **in that one plan** — an `Optional+Computed` attribute absent from the config keeps its state value in the diff. The apply reconciles both, so the state is never wrong; only that intermediate plan is incomplete. Neither `SetNew` nor `SetNewComputed` can fix it (both reject a nested path — `checkKey` looks the key up in the root schema map), and rebuilding the whole `os_disk` block in `CustomizeDiff` would risk wiping the Computed siblings that `setVMInstanceOSDisk` exists to protect. Documented at the call site.

## Tests

New non-complacent coverage, with the mutations exercised:

| Mutation applied to the code | Result |
| --- | --- |
| Remove the deprecated fallback | **7 sub-tests RED** |
| Absence decodes to `0` (the pre-fix silent behaviour) | **8 sub-tests RED** |
| Drop the divergence guard | **3 sub-tests RED** |
| Drop the os_disk both-spellings guard | `TestOSDiskBothSpellingsRejected` **RED** |
| Only the current spelling drives the grow rules | `TestOSDiskGrowRuleAppliesToEitherSpelling/deprecated` **RED** |
| Flatten stops writing the deprecated leaf | `TestSetVMInstanceOSDisk` **RED** |

All green once restored.

The os_disk tests drive the **real** `CustomizeDiff` through `Resource.Diff`, with the config wired onto the prior state's `RawConfig` — which is how `schemaMap.Diff` propagates it (`result.RawConfig = s.RawConfig`) and what `PlanResourceChange` does in production. Wiring it only as a `ResourceConfig` leaves `GetRawConfig()` null and silently skips every rule under test; the first iteration of these tests did exactly that and passed vacuously.

Writing them also surfaced a SDK behaviour worth knowing: when a diff requires replacement, `schemaMap.Diff` **replays `CustomizeDiff` with no state at all**, turning an update case into a create case. The test state has to carry the ForceNew `os_network_adapter` or the assertions test the wrong branch.

Eight existing fixtures turned out to omit a field the API contract marks `required` — they decoded to 0 silently before this change. Completed, using the current spelling, so the suite now exercises both worlds: the historical fixtures keep the deprecated names and prove the fallback.

`go install`, `go vet`, `gofmt`, and the unit suites are green. The `internal/{client,provider}/tests` acceptance suites fail on missing `SHIVA_TEST_*` credentials — a `TestMain` guard, unrelated to this change.

## Process exception — adversarial reviewer unavailable

Per `WORKFLOW_ENGINEERING.md` §3 ("No bypass"), recording this explicitly: the PLAN review could not be obtained. `codex exec` aborted with *"You hit your spend cap set by the owner of your workspace"*. A local review was done instead, and the two design decisions above were resolved by reading the SDK source rather than by assumption. **This PR has had no independent adversarial review — the human review should weigh that.** The decision stays reversible: nothing is merged to `main`, and the RC validation step still applies.


pbesret:
Moved out of the v1.12.0 train.

This pull request now targets `rc/v1.13.0` instead of `rc/v1.12.0`, and it carries the `v1.13.0` milestone along with #524 and #525.

Reason: the upstream capacity rename (`*Gb`/`*Mb` → `*Gib`/`*Mib`) has **not shipped to production** yet. Releasing the provider side first would rename 13 read attributes against an API that still answers with the old names everywhere, so the change would land ahead of what it describes. The old names are scheduled for removal on 2026-12-09, which leaves room to ship this in the next train.

Nothing about the change itself is in question — it is a scheduling decision. The branch and the review state are untouched.

pbesret:
Rebased onto the refreshed `rc/v1.13.0` (`55ffa41`, which now carries the v1.12.0 train and the v1.12.1 hotfix, #536). Same two commits, new SHAs: `8c3447e` → `2d17fc4` (feat), `6f46eb7` → `cfb4cf9` (docs). Pushed with `--force-with-lease` pinned on `6f46eb7`.

**Textual conflicts (2), resolved:**
- `internal/provider/resource_public_cloud_vm_instance.go` — `customizeVMInstanceDiff`: #533 split it into a production wrapper plus an injectable `customizeVMInstanceDiffWith(conflictOf)` that runs the plan-time IPAM collision hook first; this PR rewrote the `os_disk` size rules for the two spellings (`osDiskSizeDeclaredInRawConfig`, `both()`/`any()`, `osDiskChangedSizeAttr`). Merged as: the #533 structure (hook first, injectable), with this PR's `os_disk` logic inside the closure. No rule was dropped on either side.
- `CHANGELOG.md` — this PR's section was still titled `# 1.12.0 (Unreleased)` (written before 1.12.0 shipped). Renamed to `# 1.13.0 (Unreleased)` and placed above `# 1.12.1`; the single `latest` badge stays on 1.12.1 until the 1.13.0 release; the duplicated `# 1.11.0` header from the old hunk was dropped.

**Semantic conflicts (2), fixed in the feat commit (fixup):** two test files that came from `main` still used the field names this PR renames — `st.MinSizeGb`/`st.MaxSizeGb` → `MinSizeGib`/`MaxSizeGib` in `internal/client/vpc_vm_create_live_probe_test.go` (1.12.0 live probe), and `RAMGb` → `RAMGib` in `TestCreateVMInstanceIPAMCollisionGate` (`resource_public_cloud_vm_instance_unit_test.go`, from #533). `go vet ./...` caught both.

**Checks on the rebased branch (Go 1.24):** build OK, `gofmt` clean, `go vet ./...` OK, `staticcheck` 2025.1.1 OK, `scripts/coverage_ratchet.sh` OK (client 42.9 % ≥ 24.8, provider 28.8 % ≥ 14.3, helpers 71.7 % ≥ 70.9, race on — the schema snapshot test included), `go generate ./...` idempotent (the regenerated docs carry both this PR's GiB/MiB wording and the #533 `ip_address` description). `git range-diff 1b9da46..6f46eb7 origin/rc/v1.13.0..HEAD` shows no change beyond the resolutions above.

This is a rebase only; the review of the feature itself continues as before.

Issue #527, opened 2026-09-14
compute_virtual_machine: a create failing after the deploy activity started leaves the virtual machine orphaned outside the state


pbesret:
## Summary

On `cloudtemple_compute_virtual_machine`, when the create fails **after** the platform accepted the deployment request, the virtual machine is created on the platform but **never recorded in the Terraform state**. The user can then neither `terraform apply` nor `terraform destroy` it: the only way out is to delete the VM by hand in the console.

Reported for the marketplace deployment path (`marketplace_item_id`), but the same structural weakness affects all four create modes of the resource (clone, content library, marketplace, from scratch).

## Impact

- A VM that exists, is billable, and is invisible to Terraform.
- No `terraform destroy`, no `terraform apply` convergence — manual console cleanup only.
- Observed most often when several VMs are created in the same apply.

### D1 — the create path has no identity-recovery on failure

`POST` of the deployment returns `202` + a `Location` header carrying an activity id. **From that instant a VM may exist platform-side.** The provider then waits for the activity and, on any failure, returns an error with an empty resource id:

```go
activity, err := c.Activity().WaitForCompletion(ctx, activityId, getWaiterOptions(ctx))
setIdFromActivityState(d, activity)
if err != nil {
    return diag.Errorf("failed to deploy marketplace item: %s", err)
}
```

`setIdFromActivityState` reads **only** `state.<name>.result`. Measured against the API, `result` is populated **only when the activity reaches `completed`** — a running (or frozen) activity has no `result`. So the helper recovers nothing, `d.Id()` stays empty, and SDKv2 records nothing: the VM is orphaned.

### D2 — transport timeouts are misclassified as permanent

`retryableTransportError` treats **every** `*url.Error` whose `Timeout()` is true as non-retryable:

```go
var urlErr *url.Error
if errors.As(err, &urlErr) {
    // A configured per-request timeout / deadline reports Timeout() == true and
    // must not be retried; any other transport failure is transient.
    return !urlErr.Timeout()
}
```

The intent was to not retry the *configured per-request deadline*. But a **DNS resolution timeout** (`net.DNSError{IsTimeout: true}`) and a **dial / TLS-handshake timeout** also report `Timeout() == true`, while being ordinary transient network conditions. Measured:

| transport error | `retryableTransportError` |
| --- | --- |
| `dial tcp: lookup <host>: i/o timeout` (DNS) | `false` ← wrong |
| dial timeout | `false` ← wrong |
[...]

Issue #490, opened 2026-07-16
Terraform provider v1.9.0 does not detect a manual Cloud Temple console change to an object storage bucket's `access_type`


jschaul:
**Describe the bug**
Terraform provider v1.9.0 does not detect a manual Cloud Temple console change to an object storage bucket's `access_type`.

I have a bucket that Terraform knows as:

```json
{
  "id": "some-bucket",
  "name": "some-bucket",
  "access_type": "custom"
}
```

In the Cloud Temple console I manually changed the bucket access from `custom` to `private`.
After that, `terraform plan` still showed no diff and Terraform did not reconcile the change.

Environment
- Cloud Temple Terraform provider: 1.9.0
- Provider repository: https://github.com/Cloud-Temple/terraform-provider-cloudtemple
- Resource: `cloudtemple_object_storage_bucket`

**Current Behavior**
There is drift from actual configuration and the terraform state, it does not reconcile. Terraform shows no diff. The bucket remains recorded in state with the old `access_type = "custom"` value.

**Expected behavior**
Manual changes to the access type should get reverted on `terraform apply`. Terraform should detect that the remote bucket now has `access_type = "private"` and show a diff.

**Steps to reproduce**

1. Create or import an object storage bucket with `access_type = "custom"`.
2. Change the bucket access type in the Cloud Temple console to `private`.
3. Run `terraform plan` or refresh state.

**Additional context**

looking in the code is consistent with the bug: the bucket read
  path flattens state via internal/provider/helpers/
  helper_object_storage_bucket.go, but that flattener does not include
  access_type, and internal/provider/resource_object_storage_bucket.go
  writes that flattened map back into state on read. That makes a console-
  side custom -> private change invisible to Terraform refresh.

pbesret:
Hi @jschaul,

Thanks for the report and the accurate root-cause analysis — much appreciated.

Confirmed on our side: `access_type` (and `whitelist`) are never read back from the API, so a console-side change stays invisible to `terraform plan`. The gap actually spans two layers — the `Bucket` client struct doesn't deserialize the field, and `FlattenBucket` doesn't emit it — so the refreshed state keeps the old value.

We'll fix this (proper read-back and drift detection for `access_type` and `whitelist`) and it's scheduled for the upcoming **v1.11.0** release.

Thanks again for helping us improve the provider!

Issue #415, opened 2026-06-30
E0-7 public_cloud_vm: waiter robustness — bounded transient-404 retry + fallback GET (anti-orphan)


pbesret:
**Epic:** Refs #409  ·  **Lot:** S8 Public Cloud VM  ·  **Plan:** `E0-7`  ·  **Milestone:** v1.10.0

## Objective
MANDATORY before any resource Create: (a) bounded retry on transient activity 404 during indexing (the existing single-404 tolerance is insufficient); (b) fallback GET of the resource after a soft timeout / pending activity to recover the created ID and **prevent orphaning a live VM outside state**.

## Reused building blocks (no duplication)
Extend `internal/client/activity.go` via a **non-breaking** `WaiterOptions` field (default = current behavior for existing callers). PLAN Codex + RED/GREEN mutation tests.

## Acceptance criteria (specific)
- Mutation proof: without fallback → orphan (RED); with → GREEN.
- Existing callers unaffected.

_Plus the generic gates — see the EPIC issue._

## Notes
Spec hazards L1681-1682. Codex round-1 BLOCKING.

PR #528, opened 2026-09-14
fix(#527): never orphan a virtual machine when an async create fails

pbesret:
Refs #527

## What this fixes

A `cloudtemple_compute_virtual_machine` create that fails **after** the platform accepted the request left the virtual machine outside the Terraform state: it existed, it was billable, and neither `terraform apply` nor `terraform destroy` could reach it. Only a manual deletion in the console remained.

## Root cause

The create answers `202` + an activity id — from that instant a virtual machine may exist. `setIdFromActivityState` reads only `state.<name>.result`, which the API populates **only** when the activity reaches `completed`. Any failure during the wait therefore returned an error with an empty resource id, and SDKv2 then records nothing: `ResourceData.State()` returns `nil` on an empty id, while `Apply` returns that state even alongside error diagnostics (verified in `helper/schema` v2.35.0).

A second, aggravating defect: `retryableTransportError` treated **every** `*url.Error` whose `Timeout()` is true as permanent. A DNS resolution timeout (`net.DNSError{IsTimeout: true}`) and a dial timeout report `Timeout() == true` while being ordinary transient conditions, so a sub-second blip aborted a long activity poll outright — bypassing the bounded `maxActivityReadRetries` budget that exists for this class of failure, and treating a DNS hiccup more harshly than a connection reset, which was already retried.

## Evidence

Measured live against the dev API, polling a marketplace VMware deployment every 4 s:

- `state.running` = `{startDate, progression}` — **no `result` field**;
- `concernedItems` at submission: `marketplace_item`, `host_cluster`, `datacenter`, `datastore`;
- **~8 s in, still `running`, the API adds `{"type": "virtual_machine", "id": "<vmId>"}`**;
- at `completed` (~60 s later), `state.completed.result` carries **that same id**;
- the created virtual machine's `name` is the requested name **verbatim** (no normalisation, no suffix) and its `template` flag is `false`.

So the platform publishes the created object's identity long before completion — the provider simply never read it on the failure path.

## The fix

**F1 — never orphan the virtual machine.** A new `recoverVMCreateFailure` owns the failure path of all four create modes (clone, content library, marketplace item, from scratch). Adopting an id is **state-critical**: SDKv2 persists it as TAINTED, so the next apply DESTROYS what it points at. Adoption therefore requires strict positive evidence, and every ambiguous case fails closed with a diagnostic naming the activity, the requested name and the `terraform import` command:

- exactly **one** `virtual_machine` concerned item, minus the ids known to be something else — on the clone path, the **source** virtual machine. Zero or several candidates: no adoption. The exclusion makes the clone path correct whether or not the platform lists the source;
- a read-back proving the object **exists**, is **not a template**, and carries the **requested name**. A marketplace or content-library deployment clones a vCenter template: adopting that template would taint it and destroy it on the next apply;
- a **definitive 404** does not adopt — `confirmVMwareDeviceOrKeep` fails closed and never auto-removes, so recording a dead id would error every later refresh until `terraform state rm`;
- an **inconclusive** read-back (transport / 5xx) **does** adopt: the reported failure mode is precisely a virtual machine that outlives the apply, and a read that proves nothing must never orphan it.

`machine_manager_id` is recorded alongside the id so a later refresh can resolve the resource through the scoped listing.

**F2 — narrow the transport classification.** DNS and connection-level (dial, read, write) timeouts are transient and retried within the existing bounded budget; the configured per-request deadline stays permanent, preserving the intent of #339. Blast radius is nil for writes: only GETs and the idempotent auth POST go through `doWithRetry`. The TLS handshake timeout is deliberately left permanent — `net/http` reports it through an unexported type carrying neither `*net.OpError` nor `*net.DNSError`, so it cannot be told apart from the configured deadline by type, and leaving it out keeps the classifier fail-closed.

## Test evidence — mutation-proven

Eight mutations, eight RED, baseline GREEN:

| Mutation | Test that reds |
| --- | --- |
| restore `return !urlErr.Timeout()` | transport classification |
| revert the marketplace branch to a bare `diag.Errorf` | AST wiring guard |
| first-match instead of exactly-one | adoption |
| stop recording `machine_manager_id` | adoption |
| adopt on a definitive 404 | adoption |
| drop the clone-source exclusion | AST clone guard |
| adopt a template | adoption |
| drop the name check | adoption |

The two **AST guards** exist because the behavioural tests exercise `recoverVMCreateFailure` directly and would stay green if a call site were reverted — that revert is exactly the defect being fixed. The first guard asserts every create branch routes its failure through the recovery (and pins the count of create modes, so it can never pass vacuously); the second asserts the clone call site passes its exclusion id.

Coverage ratchet: `internal/provider` 25.2% → 26.5%, `internal/client` 40.6% → 40.7%. `go generate` produces no diff (no schema change).

## Known residual

The from-scratch mode resolves its id through `setIdFromActivityConcernedItems` (first match) **before** the error check, so the recovery honours that id without applying the template / name guards to it. That is pre-existing behaviour, on a path that has no source virtual machine, and it was left untouched rather than risk regressing the working success path.

## Out of scope

The activity freezing platform-side is the root cause of the original report and is **not** fixable in the provider; it is routed to the platform team as feedback. `cloudtemple_compute_iaas_opensource_virtual_machine` has the same uncovered hole (`setIdFromActivityState` followed by a bare `diag.Errorf`, with no audit hint). `cloudtemple_public_cloud_vm_instance` already fails closed WITH an explicit orphan-audit diagnostic and deliberately refuses to guess an id, so it is less exposed, but it still does not recover the identity the activity publishes while running. Both are tracked separately.

## Review status

⚠️ The mandatory adversarial PLAN review could not be obtained: the configured reviewer returned `Your workspace is out of credits`. Per `WORKFLOW_ENGINEERING.md` §3 ("No bypass"), a local review was performed and the exception is recorded; the decision stays reversible and **the merge gate is held** pending the reviewer's verdict and the human GO.


pbesret:
Awaiting independent reviewer

The mandatory adversarial review could not be obtained: the configured reviewer returned `Your workspace is out of credits`. Per `WORKFLOW_ENGINEERING.md` §3 ("No bypass"), a local review was performed and the exception is recorded. The decision stays reversible — nothing is merged — and the merge gate is held pending the reviewer's verdict and the human GO.

The review prompt is prepared and can be replayed as-is once credits are restored. It enumerates ten attack questions, of which four were answered by direct measurement rather than argument:

- **SDKv2 contract** — verified in `helper/schema` v2.35.0: `Apply` returns `data.State()` alongside error diagnostics, and `State()` returns `nil` if and only if the id is empty. An empty id on a create error therefore records nothing; a set id persists the resource as tainted.
- **F2 blast radius** — `doRequest` routes only GETs through `doWithRetry`; writes go through `doRequestOnce`. Making DNS and dial timeouts retryable cannot replay a non-idempotent write.
- **Wrong adoption** — the live measurement showed the marketplace deploy activity carries exactly one `virtual_machine` concerned item, and that the created VM's name matches the request verbatim with `template` false. Both were turned into hard adoption gates.
- **Partially-created VM** — `computeVirtualMachineRead` iterates `os_disk` / `os_network_adapter` from the *state*, which is empty on a recovered resource, so it performs no per-disk or per-NIC read; the half-built VM reported in the field reads and deletes normally.

Two points a reviewer should attack hardest:

1. the **inconclusive read-back adopts** branch — it prefers a possibly-dead state entry (recoverable with `terraform state rm`) over a possibly-silent orphan (recoverable only by hand). That trade-off is deliberate but debatable;
2. the **known residual** on the from-scratch mode, documented in the description.

pbesret:
## Known limitation of the evidence, stated plainly

The measurement behind this fix was taken on a deployment that **succeeded**: the `virtual_machine` concerned item appears ~8 s after the POST, while the activity is still `running`. It has **not** been measured on a **stalled** activity — which is the exact failure mode this change targets. Reviewers should weigh that.

The inference bridging the gap:

- a stalled activity is a `running` activity that never leaves that state — the same state in which the concerned item was observed, very early;
- in the internal reproduction the machines WERE materialised (present in vCenter with a UUID, one "Reconfigure" logged) before the stall, so materialisation — the moment the concerned item was seen to be written — had completed;
- the two cases partition cleanly: a stall **before** materialisation means no machine exists, nothing is orphaned and refusing to adopt is correct; a stall **after** means the concerned item was already written;
- corroborating detail: the `network_adapter` concerned item appears much later (poll 15 vs poll 2). In the reproduction the adapter was never connected, so that one would indeed be missing — the virtual machine one would not.

What would break the inference: the machine being materialised while its concerned item is written by a distinct later step that also stalled. Not excluded.

**Graceful degradation.** If the signal is absent, nothing is adopted and the error names the activity, the requested name and the import command. That is strictly better than the bare error it replaces, so the change is monotone — but the headline outcome (`terraform destroy` works unaided) is conditional on the inference above.

**Why there is no fallback.** Adding one was considered and rejected on a concrete counter-example. Any alternative is a GLOBAL lookup, and none is sound, because nothing on a virtual machine records which activity created it. The differential that looks like it solves this — census the name before the POST, adopt what is new afterwards — fails precisely under the concurrency that triggers this bug: five instances created in parallel under one name, instance A censuses zero, instance B's machine materialises, A fails and finds exactly ONE machine new since its census (B's), adopts it, taints it, and the next apply destroys a machine B also owns. An "exactly one candidate" rule does not help — the race yields exactly one candidate, the wrong one. The rationale is recorded in the file header (04d3779).

**How to close it.** Activity records outlive the machines they created. Reading a stalled create activity from a tenant that has one would settle the question in a single call, and either removes this caveat or tells us the recovery needs a different signal.

**Upstream note (not a request in this PR).** Mature providers do not have this class of bug because their create returns the object's identity synchronously — `aws_instance` sets the id from `RunInstances` before waiting for `running`. Here the create returns a `Location` to an *activity*, never the object id, so the provider structurally cannot do the standard thing and everything in this PR is compensation. The clean fix is for the create endpoint to return the created machine's id.

pbesret:
## Verdict: REQUEST-CHANGES (comment)

Adversarial review by an independent model (Claude Fable 5, read-only, maximum
effort), then both findings re-verified by the pilot before publishing. `go build`,
`go vet` and the three unit suites are green on the branch tip — neither finding is a
build or test failure. Both are reachability defects that the existing tests cannot
see.

---

### 1. D2 never runs for the incident's own error — `internal/client/api.go:608`

`retryableTransportError` evaluates

```go
if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
    return false
}
```

**before** it reaches `networkStackTimeout(err)`. In Go ≥ 1.23 the field error of this
ticket satisfies that first guard: when the dialer's own deadline fires during
resolution, `net` builds `DNSError{UnwrapErr: errTimeout}` and `errTimeout.Is` matches
`context.DeadlineExceeded` (`net.go:627-629`, `lookup.go:357`, `dial.go:562/665`). The
production transport is that dialer — `cleanhttp.DefaultPooledTransport` sets
`Dialer.Timeout: 30s`.

Reproduced through the real `net/http` + `net` stack, with a resolver blocking until
the dialer deadline fires:

```
message                                  : Get "https://…/iam/v2/auth/personal_access_token":
                                           dial tcp: lookup … : i/o timeout
as *net.DNSError                         : IsTimeout=true  Unwrap=i/o timeout
errors.Is(err, context.DeadlineExceeded) = true      <- first guard fires
networkStackTimeout(err)                 = true      <- the fix, never reached
```

So `retryableTransportError` still returns **false**, `isTransientAPIError` is false,
and the waiter (`activity.go:261`) still aborts on the first DNS blip — the exact
behaviour D2 was written to remove.

**Why the tests did not catch it.** `dnsTimeoutURLError`
(`api_resilience_test.go:149-159`) builds `net.DNSError{Err: "i/o timeout",
IsTimeout: true}` **without `UnwrapErr`**, so the synthetic error does not satisfy the
first guard and the assertion passes. The "dial timeout is transient" case uses a
private `timeoutErr{}` rather than `net`'s own `errTimeout`. This is the same trap
already recorded on this file for #490: a synthetic error that cannot distinguish a
configured timeout from a transient network one.

Consequence beyond the code: `CHANGELOG.md:8` states that "a DNS resolution timeout
or a connection-level (dial, read, write) timeout is now retried". That is not true
for the two shapes it names.

**D1 is unaffected** — a permanent failure still routes through
`recoverVMCreateFailure`, so the orphan fix itself stands.

Direction (not imposed): evaluate `networkStackTimeout` before the generic
`context.DeadlineExceeded` guard, and build the test error through the real stack
rather than by hand. A genuine parent-context deadline remains stopped by
`ctx.Done()` in `waitBeforeRetry`/`retry.Do`, and `http.Client.Timeout` is replaced by
net/http's own `httpError`, which carries neither `OpError` nor `DNSError`.

---

### 2. A single 404 is taken as proof of absence, and the advice recreates the damage — `compute_virtual_machine_create_recovery.go:176-182`

```
case readErr == nil && vm == nil:
    // Definitive absence: the platform rolled the creation back.
    … "Nothing was recorded in the Terraform state; re-run `terraform apply`."
```

One `GET` answering not-found is treated as proof the creation was rolled back. The
project's own Read path for this very endpoint refuses that inference: it confirms
absence with a scoped `ListStrict` before dropping anything
(`resource_compute_virtual_machine.go:1281-1300`, #384/#281), and `api.go:796-806`
(#415) documents GET-by-id eventual-consistency windows right after a write on this
platform.

The sequence that hurts: the `virtual_machine` item is published ~8 s after the POST
(this PR's own measurement); the wait fails just after — finding 1 makes that a DNS
blip; the read-back answers 404 because the inventory has not indexed the object yet;
the recovery records nothing and tells the operator to **re-run `terraform apply`**.
The virtual machine then materialises as an orphan, and the re-apply creates a second
one. That is the ITSM 56211 damage, produced by the fix meant to prevent it.

Note the inconsistency inside the file: every other no-adoption branch says
"audit … before re-applying". This one alone says "re-run apply".

What is proven: the decision rule, the contradiction with the codebase's own evidence
standard, and doctrine §5 (assuming an external object is gone requires strict
positive evidence). What is **not** provable from code alone: the existence and
duration of a 404 window on `GET /compute/v1/vcenters/virtual_machines/{id}` — worth
adjudicating with that in mind.

Direction (not imposed): a bounded re-check of the 404 and/or corroboration by a
scoped listing, and never advise an unaudited re-apply after a create failure.

---

### Attacks that found nothing

Wrong-id adoption (the clone source is excluded by id before any read; no name or
census fallback exists), the 403/transport → inconclusive → adopt asymmetry, the
success path (only `if err != nil` bodies changed), SDKv2 ordering of `SetId` versus
the error return (tainted as claimed), and the guard tests — five mutations on a
scratch copy all went RED, so the wiring and exclusion guards are real, not vacuous.

The unbounded wait on a perpetually running activity is pre-existing and unchanged by
this PR.

pbesret:
## READY FOR RE-REVIEW

Both blocking findings are addressed in `9794d9c`. Neither was a test failure — the branch built, vetted and passed its own suites while neither fix did what it claimed — so the work was as much about the tests as about the code.

### 1. `networkStackTimeout` is now classified before the context guards

`retryableTransportError` evaluated `errors.Is(err, context.DeadlineExceeded)` first, and since Go 1.23 a dialer-deadline DNS/connect timeout unwraps to `net.errTimeout`, which that check matches. The ticket's own error was therefore classified permanent and the retry never happened.

Safe to reorder: a genuine parent-context deadline carries neither `*net.OpError` nor `*net.DNSError` and falls through; callers still stop on `ctx.Done()`; and `http.Client.Timeout` is replaced by net/http's own `*httpError`, which carries neither type. The tail of the function simplifies — a timeout reaching it can no longer be a network-stack one.

**The test is the important half.** The old one built the error by hand and omitted the `UnwrapErr` the real stack sets, so it never exercised the guard that was short-circuiting the fix. The new one produces the error through the real `net/http` and `net` stack, with a resolver blocking until the dialer deadline fires, and pins the shape explicitly so a future Go release cannot make it pass for the wrong reason.

Restoring the old ordering makes the **new test fail and the old one still pass** — the whole problem in one observation.

### 2. An absence is now confirmed before it is believed

The adoption rule is unchanged: adopting an id that no longer resolves really would poison every later refresh, which is why case 4 exists. What changes is that a single not-found no longer counts as proof.

Two further read-backs, by id, no listing, on a path that has already failed. Anything contradicting the absence wins immediately — a virtual machine that appears is adopted, a failing read or a cancelled context is inconclusive and adopted rather than orphaned. Only a run of not-founds is believed, and the diagnostic then asks the operator to **confirm before re-applying**, like every other branch that records nothing. The `re-run terraform apply` instruction is gone: it was the one advice able to manufacture the incident being fixed.

The backoff is a var so tests drive the path without sleeping; the pre-existing 404 case uses it too, so the suite stays under a second.

### Also corrected

Both CHANGELOG entries. The first asserted a retry that did not happen; the second did not say that an absence is verified rather than assumed.

### Proof

Nine mutations across the two fixes, every one caught: removing the confirmation, swallowing the re-read error, ignoring context cancellation, restoring the `re-run apply` advice, and restoring the original guard ordering. `go build`, `go vet` and the three unit suites are green.

pbesret:
## Re-review — VERDICT: GO, no blocking finding

Independent adversarial re-review (Claude Fable 5, read-only, maximum effort) of
`9794d9c`, scoped to the two blocking findings and — above all — to whether the fixes
broke anything. Build, vet and the three suites are green, including `-race`.

### The two findings are answered

**Ordering.** `networkStackTimeout` is evaluated first (`api.go:629-631`). The
incident's shape comes from the *dialer's own* `Timeout`, which `net` implements as a
child-context deadline: when it fires during the lookup, `net` builds
`DNSError{IsTimeout: true, UnwrapErr: errTimeout}`, and `errTimeout.Is` matches
`context.DeadlineExceeded`. Restoring the old ordering fails **only** the new
real-stack test; the hand-built one still passes — the blind spot, demonstrated.

**Absence.** A first not-found now routes through `confirmVMAbsence`
(`recovery.go:227-230`), and the confirmed-absence diagnostic no longer sends the
operator straight back to apply (`:241-244`).

### The risk introduced by the fix does not exist — and the reason is better than the one I gave

I had flagged that a parent-context deadline firing *during a dial* would now be
reclassified retryable, and justified the fix by saying callers stop on `ctx.Done()`.

That premise is wrong, and the reviewer showed why: **`http.Transport` detaches the
dial from the request context** (`context.WithoutCancel` in `dialConnFor`), and
`getConn` returns the request's own ctx error. A parent deadline or cancellation
therefore never surfaces as a `*net.OpError` / `*net.DNSError` at all.

Measured through the real stack, on both sides:

| Case | Error | `networkStackTimeout` | Class |
|---|---|---|---|
| Dialer deadline (the incident) | `dial tcp: lookup …: i/o timeout` | **true** | retryable |
| Parent context deadline | `context deadline exceeded` | false | permanent |
| Parent context cancel | `context canceled` | false | permanent |

Cancellation could not match in any case: `errCanceled` has no `Timeout()`, and
`newDNSError` sets `IsTimeout` only from `net.Error.Timeout()`.

Callers stopping on `ctx.Done()` was verified as a second line of defence, not as the
reason: probes driving `doWithRetry`, `waitForActivityCompletion` and
`waitForBackupJobCompletion` with an already-expired parent context made **one call
each, ~152 ms elapsed** — no spin, no stall. `4e73cea` replaces the comment's weaker
justification with this measured structural one, so a future reader does not think the
ordering is a trade-off held in place by the call sites.

The configured per-request deadline stays permanent: `Client.Timeout` is a child
deadline whose failure net/http replaces with its own timeout error, carrying neither
type.

### Everything else checked

`confirmVMAbsence`'s four branches are correct (VM appears → adopt; read error →
inconclusive → adopt; `ctx.Done()` → inconclusive → adopt; only a run of not-founds is
believed). Cost is bounded at ≤ 6 s and two by-id reads, well inside the SDKv2 20-minute
create deadline. The `vmAbsenceBackoff` var is unexported, reassigned only by tests,
and the provider suite runs no `t.Parallel`. No other error shape changes class — TLS
handshake timeouts, proxy CONNECT errors, HTTP/2 stream errors and `io.EOF` are all
unaffected.

Residual trade-off, stated rather than hidden: a genuine absence combined with a blip
on a confirmation read adopts a dead id, whose refresh then fails closed until
`terraform state rm`. That is the pre-existing case-3 behaviour, already cleared in the
first review — recoverable and non-destructive, unlike the orphan it replaces.

### Tests

Hermetic — `.invalid` plus a resolver whose `Dial` never opens a socket, and `httptest`
on loopback. No DNS, no network. Client suite green under `-race`; the two timer-based
tests green at `-count=15 -race`; the cancelled-context subtest green at `-count=3000`
both with `-race` and at `GOMAXPROCS=1`. All nine mutations caught.

The CHANGELOG entries now match the code on both points.

pbesret:
Conflict resolved in `6e47e89`.

#530 landed first, so both branches carried a `# 1.12.0 (Unreleased)` header. That was the **only** conflict — the code merged cleanly, including `resource_compute_virtual_machine.go`, which both changes touch.

Resolved by **keeping both sides** under one header, in the house order (UPGRADE NOTES, ENHANCEMENTS, BUG FIXES). Nothing was chosen over anything: the eight entries are #530's two upgrade notes and four enhancements, plus this branch's two bug fixes, each carried across unchanged.

Verified after the merge rather than assumed: build, vet and the three unit suites green, one test from each side run explicitly to confirm the two feature sets coexist, and `go generate` a no-op.

The re-review verdict above (GO, no blocking finding) stands — the merge changed no code on this branch.

Issue #529, opened 2026-09-14
OpenIaaS and Public Cloud VM: recover the virtual machine identity on an async create failure (follow-up to #527)


pbesret:
Follow-up to #527, which fixed the orphan-on-create-failure hole on `cloudtemple_compute_virtual_machine` only. The other two asynchronous virtual machine resources are exposed to the same class of defect, to different degrees.

## Why this is a real hole

An asynchronous create answers `202` + an activity id: from that instant an object may exist platform-side. Returning the wait failure with an empty resource id makes SDKv2 record **nothing** — `ResourceData.State()` returns `nil` on an empty id — so the object is orphaned: unreachable by `terraform apply` and `terraform destroy`, and only removable by hand.

#527 established that the platform publishes the created object in the activity's `concernedItems` **while the activity is still running**, well before `state.result` exists. That is an authoritative identity the provider can adopt instead of orphaning the object.

## `cloudtemple_compute_iaas_opensource_virtual_machine` — same uncovered hole

Both create branches (from template, and from a marketplace item) do:

```go
activity, err := c.Activity().WaitForCompletion(ctx, activityId, getWaiterOptions(ctx))
setIdFromActivityState(d, activity)
if err != nil {
    return diag.Errorf("failed to create virtual machine, %s", err)
}
```

`setIdFromActivityState` only reads `state.<name>.result`, absent until the activity completes. So a stalled activity, a transient network error, or a cancelled apply returns an empty id and orphans the VM — with **no audit hint** in the diagnostic telling the operator what to look for.

## `cloudtemple_public_cloud_vm_instance` — mitigated, not solved

`createVMInstanceWith` is already state-safety aware: on a wait failure it fails closed with an explicit diagnostic ("If a VM was created it is now ORPHANED outside the state — audit … and import it or delete it before re-applying"), and it deliberately refuses to guess an id from the name. That is meaningfully better than the VMware resource was.

It still does not **recover** the identity: the id is taken only from the *completed* activity (`concernedItems` of type `vmi`, else `state.result`), so a create that never completes leaves the operator to find and import the VM manually.

## Work

1. **Measure first.** Confirm, on the VM Instances API and on the OpenIaaS API, whether the create activity publishes its `concernedItems` (`vmi` / `virtual_machine`) **while running**, as the VMware marketplace activity does. #527's design rests on that measurement; it must not be assumed for another API.
[...]

PR #532, opened 2026-09-16
Release v1.12.0 — VPC static IP on the inline os_network_adapter, and no more orphaned VM on a failed async create

### LINKED ISSUE #376: feat(vpc): create a VM with its inline os_network_adapter on a VPC network (VM Instances, OpenIaaS, VMware)

### LINKED ISSUE #527: compute_virtual_machine: a create failing after the deploy activity started leaves the virtual machine orphaned outside the state

pbesret:
Closes #376
Closes #527

Release v1.12.0 — a VM can be created directly on a VPC private network with a chosen static IP, and a create that fails after deployment no longer leaves the virtual machine outside the state.

Two issues, both verified against **production** before this pull request was opened.

### `os_network_adapter` on a VPC network, with a chosen static IP (#376)

The inline block of all three VM resources now accepts a VPC-backed network, so a VPC-only machine is expressible in a single `terraform apply` — previously it took a two-step dance through a standalone adapter resource. `ip_address` registers the address as the adapter's VPC static IP; omit it and the platform assigns one.

The decision logic lives once and is shared, so the surfaces cannot drift in what they promise or how they fail. Each guard exists because of something the platform actually does:

- **the address is only honoured on a VPC-backed network** — on a plain network the platform silently discards it, so a chosen address would be dead configuration the user believes took effect;
- **two blocks cannot claim one address on the same network** — the platform assigns one address per (virtual machine, network) pair, and its refusal otherwise arrives from the create activity, leaving an untracked virtual machine behind;
- **an address already registered on the target network is refused** — measured: the platform does *not* refuse it. It creates the resource, reports success, and registers nothing, while Terraform records the requested address. Since `ip_address` is write-only, no refresh ever compares it and no plan re-enters reconciliation: the divergence is permanent and invisible. Only a pre-flight can catch it, so the check runs at plan and again before the platform call, and fails closed on a read error.

### No more orphaned virtual machine on a failed async create (#527)

Customer incident: an apply deploying from a marketplace item timed out, the virtual machine existed on the platform but was absent from the state, reachable by neither apply nor destroy — manual deletion, and a leaked 32 GiB volume.

Two defects, both fixed. The deployment request answers with an activity id, and from that moment a virtual machine may exist; the provider now recovers its identity from the activity — which publishes it as soon as the object is materialised, well before completion — and records it **tainted**, so `destroy` removes it and `apply` replaces it with no manual cleanup. Identity is adopted only on unambiguous evidence, and never the source virtual machine of a clone. Separately, a transient network timeout no longer aborts a long-running operation: a DNS or connection-level timeout is retried within the existing bounded budget, where a sub-second DNS hiccup previously failed the whole apply.

## Upgrade risks

**One RED: #376.** A configuration that sets `ip_address` on a Private Backbone adapter of `cloudtemple_public_cloud_vm_instance` will now **fail at apply**, without any change to the Terraform code. The value used to be accepted and silently discarded — it never took effect — so nothing that worked stops working, but an apply that used to pass will now stop. Remove the argument, or move the adapter to a VPC network. This is the first entry of `UPGRADE NOTES`.

The inline block remains `ForceNew`: now that VPC networks are accepted there, changing an interface's `network_id` or `ip_address` **replaces the virtual machine**. To relocate an adapter or move its address in place, manage it with a standalone `cloudtemple_*_network_adapter` resource instead.

**#527 is RED as a description of the bug, not of the fix.** The fix strictly reduces damage: where an apply used to leave an unmanageable billable machine, it now records it tainted. The one behaviour change to know about is that a virtual machine adopted this way is destroyed and recreated by the next apply — run `terraform untaint` first if you verify it is actually complete.

## Production verification

All four creation paths that can carry an inline VPC static IP were exercised against **production** and verified on the VPC plane by MAC lookup, requested address versus registered address:

| Path | Mechanism | Requested | Registered |
|---|---|---|---|
| OpenIaaS marketplace | `networkData[].ipAddress` | `172.16.0.200` | `172.16.0.200` (`source=xoa`) |
| OpenIaaS template | `networkAdapters[].ipAddress` | `10.0.15.10` | `10.0.15.10` (`source=xoa`) |
| VMware | adapter PATCH | `10.0.15.20` | `10.0.15.20` (`source=vmware`) |
| VM Instances | `networkInterfaces[].ipAddress` | `10.0.15.30` | `10.0.15.30` (`source=vmi`) |

The collision guard was proven against production too: it refused real taken addresses, naming the actual holder, and passed on free ones. Every probe was created powered off and destroyed immediately; both tenants were returned to their exact prior counts of virtual machines and static IP registrations.

**Not exercised in production**, stated so no reader assumes otherwise: the collision guard *refusing* at apply time, the fail-closed-on-read-error path, the update paths (relocation, the "already ours" exemption), the two-blocks-same-network refusal, and the deferred rollback on an error return. Those rest on the unit tests and the mutation proofs.

## Review

Both feature pull requests were reviewed by an independent adversarial model, with the roles separated per the engineering workflow.

#530 returned **GO** with no blocking finding across seven attacks. #528 returned **NO-GO** on two blocking findings — the retry fix was inert for the incident's own error shape, certified by a test that built the error by hand and omitted the field the real stack sets; and a single not-found was taken as proof of a rollback, with advice that could manufacture the very incident being fixed. Both were re-verified independently, fixed, and the re-review returned **GO**.

Nineteen mutations across the two pull requests, every one caught.

## Not done here — human gates

Merging this pull request, and creating the `v1.12.0` tag afterwards, are human decisions. Neither was performed.


pbesret:
RC-VALIDATION: OK rc/v1.12.0 2726caf6ddf3b0d4931028d1a1deca2699fe1072

Validation of the RC state before merging into `main`, per `WORKFLOW_GIT_EPIC.md`.

### Checks executed

**CI on the RC tip `2726caf`** — Build ✅, Unit tests & vet ✅, generate ✅.

**Gates re-run locally on the RC state** (worktree reset to `2726caf`, no untracked file):

| Gate | Result |
|---|---|
| `gofmt -l` (whole module) | clean, nothing to reformat |
| `go build ./...` | pass |
| `go vet ./...` (compiles every `_test.go`) | pass |
| `go test ./...` | all unit suites pass — see the note below on the two acceptance packages |
| `go generate` (docs) | no diff |
| `scripts/coverage_ratchet.sh` | `internal/client` 40.9% (floor 24.8%) · `internal/provider` 27.9% (floor 14.3%) · `internal/provider/helpers` 71.7% (floor 70.9%) |

Coverage moved up on both packages this train touched: `internal/client` 40.6 → 40.9%, `internal/provider` 26.1 → 27.9%.

**Two packages report FAIL locally and it is not a defect:** `internal/client/tests` and `internal/provider/tests` are the acceptance suites, which abort without `SHIVA_TEST_CLIENT_ID` / `SHIVA_TEST_SECRET_ID` / `SHIVA_TEST_COMPANY_ID`. They are not run in CI either — the `go test ./...` step in `test.yml` is commented out. Stated rather than hidden: this train is **not** covered by the acceptance suites, and leans instead on the unit tests, the mutation proofs and the production verification below.

**staticcheck could not be run locally**: the installed binary is built with go1.23.2 and refuses a go1.24 module. It *is* run in CI on this RC state, with a version pinned for Go 1.24, and that job is green — so the gate is covered, just not by me.

### Production verification

Both issues were verified against production before this pull request, not only on DEV.

All four creation paths that can carry an inline VPC static IP were exercised and confirmed on the VPC plane by MAC lookup — requested address equals registered address in every case, across two tenants:

| Path | Mechanism | Requested | Registered |
|---|---|---|---|
| OpenIaaS marketplace | `networkData[].ipAddress` | `172.16.0.200` | `172.16.0.200` (`source=xoa`) |
| OpenIaaS template | `networkAdapters[].ipAddress` | `10.0.15.10` | `10.0.15.10` (`source=xoa`) |
| VMware | adapter PATCH | `10.0.15.20` | `10.0.15.20` (`source=vmware`) |
| VM Instances | `networkInterfaces[].ipAddress` | `10.0.15.30` | `10.0.15.30` (`source=vmi`) |

The collision guard was proven against production as well: it refused real taken addresses naming the actual holder, and passed on free ones. Every probe was created powered off and destroyed immediately; both tenants were audited back to their exact prior counts of virtual machines and static IP registrations, with nothing added and nothing left behind.

The marketplace path deserves a note: it used to be **refused** by the provider because `NetworkDataMapping` carried no address field and the route had never been measured. The marketplace module maintainer confirmed the endpoint accepts `ipAddress` despite the published swagger omitting it — the same swagger-versus-reality gap #376 already found on the VM-create route — and the measurement above is what turned that statement into evidence.

**Not exercised in production**, so that no reader assumes otherwise: the collision guard *refusing* at apply time, the fail-closed-on-read-error path, the update paths (relocation, the "already ours" exemption, the reconciliation entry condition), the two-blocks-same-network refusal, a marketplace item with more than one adapter, and the deferred rollback on an error return. Every probe ran with `power_state = "off"`.

### TF prod breaker verification

Explicitly reviewed for both issues in the train. #527 was unclassified before this validation and has been assessed and set.

| Issue | Lot | Risk | Priority | TF prod breaker |
|---|---|---|---|---|
| #376 inline `os_network_adapter` on a VPC network | S2 VPC | `scope-gap` | P1 | **RED — peut casser le TF client** |
| #527 orphaned VM on a failed async create | S5 State-safety | `drift` | P1 | **RED — peut casser le TF client** |

**#376 is RED** because a configuration that sets `ip_address` on a Private Backbone adapter of `cloudtemple_public_cloud_vm_instance` now fails at apply with no change to the Terraform code. The value was previously accepted and silently discarded, so nothing that *worked* stops working — but an apply that passed will now stop. It is the first entry of `UPGRADE NOTES`.

**#527 is RED as a description of the bug**, not of the fix: a production client saw an apply fail leaving a billable virtual machine outside the state, manageable by neither apply nor destroy. The fix strictly reduces the damage. The one behaviour change to know is that a virtual machine adopted through the recovery is recorded **tainted**, so the next apply destroys and recreates it — `terraform untaint` keeps it if you verify it is complete.

Rationale for the newly classified #527: `drift` describes the nature of the damage — the state no longer reflects reality — rather than the trigger, which is a mishandled transient network timeout and would read as `transient-retry`. Both halves shipped together; the dominant harm is the untracked machine. P1 follows the calibration of the other client-reported RED regressions on this board (#350, #355, #395, #397); P0 would be defensible given the billing impact, and is worth a word if the release owner disagrees.

### Review trace

Roles were separated per the engineering workflow: each change was piloted by one model and reviewed by an independent adversarial one.

| Pull request | Verdict |
|---|---|
| #530 (#376) | **GO**, no blocking finding, across seven attacks |
| #528 (#527) | **NO-GO** on two blocking findings → fixed → re-review **GO** |

The #528 findings are worth recording because neither was a test failure: the branch built, vetted and passed its own suites while one of its two fixes did nothing. The retry fix was short-circuited by a guard evaluated before it, and the test certified the behaviour only because it built the error by hand, omitting the field the real Go stack sets. The second finding was a single not-found taken as proof of a rollback, with advice that could manufacture the very incident being fixed.

Nineteen mutations across the two pull requests, every one caught.

### Aggregated issues

`Closes` = 2 issues: **#376**, **#527** — both in `Awaiting release` on the board, both confirmed present in `closingIssuesReferences` on this pull request after creation.

#529 (the same recovery for OpenIaaS and Public Cloud VM) is follow-up work and is deliberately **not** in this train. #526 (#524/#525, the Public Cloud VM capacity rename) was moved to `rc/v1.13.0` because its upstream API rename has not shipped to production.

### Not done here — human gates

Merging this pull request into `main`, and creating the `v1.12.0` tag afterwards, are human decisions. Neither was performed. If the RC branch receives any commit after this marker, the marker is invalidated and the validation must be replayed on the new SHA.

One thing to check before merging: the CHANGELOG section is dated **September 16th, 2026**. Re-date it if the release happens later — the entry should say when the version actually shipped.

Alexandre-Cann (approved):

Issue #534, opened 2026-09-17
Pre-create IPAM collision gate: tolerate the reclaim window of a destroyed VM's static-IP registration (follow-up of #533)


pbesret:
## Context

Follow-up of #533. Since v1.12.1 the plan-time IPAM collision check defers to the authoritative pre-create check whenever the plan carries no identity (a fresh create, or the create half of a replacement — Terraform plans the latter against a null prior state). A destroy-before-create replacement of a virtual machine that holds a VPC static `ip_address` therefore plans correctly and, at apply, destroys the old machine first, then runs the pre-create check before creating the new one.

## The residual

The platform reclaims a destroyed virtual machine's static-IP registration **after** the delete activity completes, not always instantly. Measured during the PROD validation of #376 on the VM Instances surface: right after a destroy the per-network static-IP listing still returned the registration on the first pass, and no longer on the second. The provider calls the create immediately after the delete activity completes, and the pre-create check performs a single strict listing, so the create can be refused once with:

```
os_network_adapter (device_index 0) requests ip_address ..., which is ALREADY registered ... for virtual machine <the destroyed VM> ...
```

At that point the old machine is destroyed and removed from the state, nothing has been created, and the next `apply` converges. The outcome is fail-closed, retryable and leaves no orphan, and the diagnostic (since v1.12.1) tells the user to retry once the registration has disappeared — but it is an ugly failure for the most common replacement scenario.

## Proposed mitigation (provider side)

In the pre-create check of the three VM resources (`createVMInstanceWith`, `validateInlineAdapterIPPreconditionsOnCreate` and the OpenIaaS create paths that call `rejectInlineAdapterIPAlreadyRegistered` with an empty owner), when the address is held by a registration whose `virtualMachine` no longer exists (a strict read of that VM answers a confirmed absence, re-read once against the eventual-consistency window handled in #527), **wait, bounded**, for the registration to disappear from the listing before proceeding; refuse as today on timeout, on a read error, or when the holder VM still exists (that is the `create_before_destroy` case, which must keep failing).

Not in scope: assuming an address is free because its holder VM is gone — the create must still observe the registration disappear (a stale registration may still make the platform silently skip the registration of the new VM, the very defect the check exists for).

## Evidence needed first

[...]

PR #535, opened 2026-09-17
fix(#533): stop rejecting a VM's own VPC static IP when planning its replacement (hotfix → v1.12.1)

### LINKED ISSUE #533: VM instance replacement plan rejects its own VPC static IP reservation (v1.12.0)
## Problem

With Cloud Temple provider 1.12.0 and OpenTofu 1.12.6, adding `cloud_init` to an existing stopped `cloudtemple_public_cloud_vm_instance` triggers replacement, but planning fails because `os_network_adapter.ip_address` is already registered to the **same VM being replaced**.

The VM has one VPC adapter and an explicitly reserved private IP. Default destroy-before-create ordering is used; `create_before_destroy` is not enabled.

## Reproduction

1. Create a stopped VM on a VPC private network with `os_network_adapter.ip_address` explicitly set.
2. Keep the VM, VPC network and IP unchanged.
3. Add a minimal cloud-init payload, for example:

```hcl
cloud_init = {
  cloud_config = "#cloud-config\nhostname: control-01\n"
}
```

4. Run `tofu plan`.

Observed on the released provider, with the existing remote state: planning fails with `os_network_adapter[0] requests ip_address ... which is ALREADY registered ... registered by vmi for virtual machine <the current VM id>` and `Choose a free address`.

A subsequent `tofu plan` on the unchanged canonical configuration succeeds with exit code 0 and `No changes`.

No apply, deletion, replacement or state removal was performed. The diagnostic plan was run in a temporary copy of the configuration against the original remote backend. The first probe also changed `power_state` from `off` to `on`. A second probe adding **only** `cloud_init`, keeping `power_state = "off"`, reproduces the exact same collision error. CPU, RAM, image, AZ, network and address were unchanged in both probes.

## Expected behaviour

Planning a destroy-before-create replacement should be possible when the current resource owns the reservation and deletion releases it. A reservation belonging to another VM must still fail. Create-before-destroy with the same IP must not be accepted as a safe replacement, and the authoritative pre-create collision check must remain in place.

## Suspected cause (not yet proven by a provider unit test)

`inlineAdapterIPCollisionDiff` passes `diff.Id()` to `rejectInlineAdapterIPAlreadyRegistered`. The latter correctly exempts the resource's own reservation for an in-place operation. During the replacement diff pass, the resource ID may be empty, so its own reservation is reported as a conflict.

Sources at tag v1.12.0:
- `internal/provider/resource_public_cloud_vm_instance.go`: `cloud_init` is `ForceNew`.
- `internal/provider/vpc_inline_adapter_ip.go`: ownership check and `inlineAdapterIPCollisionDiff`.

Please cover the replacement diff lifecycle, no-op on the existing VM, a real foreign-owner collision and create-before-destroy. The initial create is working; this report concerns replacement planning.

pbesret:
Closes #533

**Direct hotfix to `main`** for a blocking planning regression reported on the released **v1.12.0**. Per the RC-flow rules this PR requires a `HOTFIX-APPROVED: #533` comment from an authorized human, a human review and a human merge; the `v1.12.1` tag is then created by a human from the merge commit on `main`, and the fix is ported into the active `rc/v1.13.0` through a dedicated PR.

## The defect

Planning the **replacement** of a `cloudtemple_public_cloud_vm_instance` that holds a VPC static `ip_address` (for example adding `cloud_init`, which is `ForceNew`) failed with:

```
os_network_adapter[0] requests ip_address ... which is ALREADY registered ... registered by vmi for virtual machine <the id of the VM being replaced> ... Choose a free address.
```

The plan-time IPAM collision check introduced in 1.12.0 excuses the resource's own registration through the resource id. Terraform (and OpenTofu) plan a replacement in **two** `PlanResourceChange` calls; the second one — the create half — is made against a **null prior state** (`internal/terraform/node_resource_abstract_instance.go`, `plan()`, `if action.IsReplace()`), and the old identity travels only in `PriorPrivate`, which terraform-plugin-sdk/v2 never exposes to `CustomizeDiff`. In that call `ResourceDiff.Id()` is `""`, the raw state is null, and a fresh create is indistinguishable from the create half of a replacement. The hook therefore reported the VM's own registration as a foreign conflict — a false FAIL, which its contract ("a false PASS is possible, a false FAIL never") forbids.

The independent review of the plan surfaced a second gap of the same family on the two Compute surfaces (`cloudtemple_compute_virtual_machine`, `cloudtemple_compute_iaas_opensource_virtual_machine`), which share the hook and the update gate: **with** an identity, ownership was proven only through the registration's `virtualMachine` link, while these surfaces register by MAC and the platform does not always link the machine (the client fixtures carry `virtualMachine: null` for `xoa`/`vmware` rows). A MAC-owned own registration would then have been refused on an in-place plan, on the first pass of a replacement plan, and by the authoritative update gate.

## The fix

- `inlineAdapterIPCollisionDiff` (shared by the three VM resources) **no longer adjudicates when the plan carries no identity** (`diff.Id() == ""`): a fresh create, or the create half of a replacement. The verdict belongs to the authoritative pre-create check, which is unchanged and correctly ordered: under the default destroy-before-create ordering it runs after the delete activity completes and re-reads the IPAM plane; under `create_before_destroy` the old VM still holds the address, so the create is refused before anything is created. With an identity the behaviour is unchanged: the own registration passes, anything else is refused — including on the **first** pass of a replacement plan, which still stops a destroy-before-create replacement onto a foreign registration before the old VM would be destroyed.
- `rejectInlineAdapterIPAlreadyRegistered` now also recognises the resource's own registration by the **MAC address of one of its own adapters** (`ownerMACs`, read from the planned `os_network_adapter` blocks — `mac_address` is Optional+Computed on the Compute surfaces and absent from the VM Instances block, whose registrations are proven to carry the machine link). The VPC plane addresses a registration by MAC (`GET /vpc/v1/static_ips/mac/{mac}`), and a MAC is unique on a network, so the match is as positive as the machine link. An empty MAC on either side never matches.
- A create-time collision refusal now carries a hint for the replacement case: keep the default destroy-before-create ordering and retry once the registration has disappeared; `create_before_destroy` cannot reuse the same address.
- `customizeVMInstanceDiff` is split into an injectable `customizeVMInstanceDiffWith(conflictOf)` plus the production wrapper (no behaviour change), so a test can drive the real resource through `Resource.SimpleDiff` — the exact entry point of the gRPC `PlanResourceChange` — with a recorded IPAM view and no network.
- `ip_address` description, regenerated docs, CHANGELOG `1.12.1` (single `latest` badge moved).

## Tests (mutation-proven)

- `vpc_inline_adapter_ip_replacement_test.go` — a protocol-faithful harness that builds what `grpc_provider.PlanResourceChange` builds (`ShimInstanceStateFromValue`, `RawState`/`RawPlan`/`RawConfig`, `NewResourceConfigShimmed` from the proposed state, `SimpleDiff`), with cty values produced from JSON against the resource's implied type, and a **recording** IPAM function so every expected pass asserts the lookups actually made (no vacuous pass). On the real `cloudtemple_public_cloud_vm_instance`: in-place own → pass (1 lookup); in-place foreign → refused (no replacement hint); IPAM error → fails closed; replacement pass 1 (identity, `cloud_init` added) → pass and `RequiresNew()`; **replacement pass 2 (null prior) → pass, 0 lookups — the #533 case**; no identity + foreign holder → deferred; offline `Resource.Diff` re-diff → pass. On the real `os_network_adapter` schemas of the two Compute surfaces: identity + machine link → pass; identity + **MAC-owned registration without machine link** → pass; identity + foreign MAC → refused; no identity → deferred whatever the holder.
- `TestCreateVMInstanceIPAMCollisionGate` — the apply-time gate on VM Instances: the address held by the VM being replaced (`create_before_destroy`, or not yet reclaimed) refuses the create before any side effect, names the holder and carries the hint; a free address passes and the create proceeds. This gate had no direct unit test on this surface until now.
- `TestRejectInlineAdapterIPAlreadyRegistered` — own-MAC passes (case-insensitive), foreign MAC refused, no evidence refused, create-time hint present, update-time hint absent.

Mutation evidence:
- early return removed (v1.12.0 behaviour): the null-prior cases fail with the exact production diagnostic (`... registered by vmi for virtual machine vm-self ...`) on VM Instances and on both Compute surfaces; every identity case stays green; restored → all green.
- MAC branch of `staticIPOwnedBy` removed: the own-MAC cases fail on the table test and on both Compute surfaces; everything else stays green; restored → all green.

CI-equivalent locally (Go 1.24): `gofmt` clean, `go vet ./...` OK, `staticcheck` 2025.1.1 OK, `scripts/coverage_ratchet.sh` OK (client 40.9 % ≥ 24.8, provider 28.6 % ≥ 14.3, helpers 71.7 % ≥ 70.9, race detector on), `go generate ./...` idempotent.

## Residual, stated

The platform reclaims a VM's static-IP registration after the VM is destroyed, **not always instantly** (measured during the PROD validation of #376: the listing still showed the registration on the first pass after a destroy). A destroy-before-create replacement calls the create right after the delete activity completes, so the pre-create gate may refuse once, naming the destroyed VM as the holder. The old VM is destroyed and out of the state, nothing is created, and the next `apply` converges: fail-closed, retryable, no orphan. The diagnostic now says so. A bounded wait in the pre-create gate while the holder VM no longer exists is tracked separately in #534; it is out of this hotfix because it changes the authoritative gate on three surfaces and needs a live measurement of the window.

## Delivery checklist

- [ ] `HOTFIX-APPROVED: #533` comment by an authorized human
- [ ] human review and approval
- [ ] human merge (merge commit or squash; no rebase merge)
- [ ] tag `v1.12.1` on the merge commit on `main` (release workflow publishes)
- [ ] port into `rc/v1.13.0` through a dedicated PR, replay the RC validation on the new SHA, update the RC release notes
- [ ] board #11: this PR added with Lot `S2 VPC`, Priority `P0`, Risk `drift`, TF prod breaker `RED` (copied from #533); after the merge, PR → `Done`, #533 → `Done`

## Review trace

Pilot: Claude (Fable 5.1). Independent adversarial reviewer per the engineering rules: Codex (gpt-5.5, `codex exec --sandbox read-only`).

- PLAN round 1 — Codex — **NO-GO**: the first design deferred, at plan time without identity, only registrations carrying a `virtualMachine` link; that predicate is unsound on the Compute surfaces (MAC-owned rows with `virtualMachine: null`). Design dropped for the one Codex recommended (no adjudication without identity).
- PLAN round 2 — Codex — **NO-GO**: with an identity, ownership was still proven only by the machine link, so a MAC-owned own registration was refused on in-place plans, the first replacement pass and the Compute update gate. Fixed by the MAC-ownership proof described above; nits on the description and CHANGELOG wording applied.
- Pre-commit round 3 on the real diff — Codex — **could not complete**: the run stopped on the workspace spend cap ("You hit your spend cap set by the owner of your workspace"). Per the rules' "no bypass" clause the review was performed by another available model, Claude Opus 5, read-only, on the same brief; its verdict is recorded verbatim in a comment below: **NO-GO** on one BLOCKING finding — the regenerated docs were one sentence behind the schema description (the description had been amended after the last `go generate`), which would have failed the CI `generate` job — plus three MINOR (comment-level: per-VM scope of the MAC excuse, a too-absolute "reclaimed by then", a too-strong "as positive as the machine link") and three NIT. The blocker was fixed by re-running `go generate ./...`; every MINOR and two NIT were applied (comments, CHANGELOG wording, no dead MAC collection on the create path); the third NIT (the hint also reaching the Compute create refusals, conditionally phrased) is accepted as is. The delta was re-submitted to the same reviewer; its answer is recorded verbatim below the first verdict. The decision stays reversible: nothing is merged, and the merge remains conditioned on the human `HOTFIX-APPROVED` and review.
- Committed-diff pass — **pending Codex** (spend cap); to be run once the cap is raised, or replaced by the human review of this PR.


pbesret:
## Independent adversarial review — pre-commit pass on the real diff (round 3)

**Context.** The project's declared reviewer (Codex, `codex exec --sandbox read-only`) delivered the two PLAN rounds (both NO-GO, both resolved — see the PR description) and then stopped mid-way through this round on the workspace spend cap (`You hit your spend cap set by the owner of your workspace`). Per the engineering rules' "no bypass" clause, this pass was performed by another available model, Claude Opus 5, read-only, on the same brief (root cause, prior findings, test and mutation evidence, seven enumerated attack questions, the full `git diff` and the new test file). Its verdict is reproduced **verbatim** below. The one BLOCKING finding (regenerated docs one sentence behind the schema description) was fixed by re-running `go generate ./...`; the MINOR/NIT findings were addressed as described in the follow-up comment. The Codex pass on the committed diff remains due once the cap is raised.

---

REVIEWER: Claude Opus 5 (fallback reviewer — Codex spend cap reached; see review trace)

**Verification performed first.** The brief's section D diff is byte-identical to `git diff` on the worktree (565 vs 564 lines, the sole difference being my extraction's trailing newline), and section E is byte-identical to the untracked `internal/provider/vpc_inline_adapter_ip_replacement_test.go`. The brief did not misrepresent the tree. The targeted suite passes, and so does the whole package (`go test ./internal/provider -count=1` → `ok`, 0.649s). `gofmt -l internal/` is empty, `go vet ./internal/...` is clean, and — resolving the brief's one pending item — pinned `staticcheck 2025.1.1` run with the 1.24 toolchain is clean over `./...`, not just the provider package. One brief claim is **false**, see A5/BLOCKING.

**A1. Is MAC equality sound as positive proof of ownership?** Yes, and the decisive argument is not the MAC's uniqueness but the *direction* of the change. `staticIPOwnedBy` (`internal/provider/vpc_inline_adapter_ip.go:192-208`) is a strict superset of the v1.12.0 predicate: its first branch (`:196`) adds `holder.VirtualMachine.ID != ""`, which is a no-op because every caller guards on `ownerVMID != ""` (`:176`, `:582` via `:537`), so branch one is exactly the old condition and the MAC branch (`:199-207`) only ever *adds* passes. The widened predicate therefore cannot manufacture a new false FAIL anywhere — which settles the brief's sub-question about a MAC changed in the same apply. Planned is the NEW MAC (`plannedInlineAdapters` is `d.Get("os_network_adapter")`, `:772-775`, the Optional+Computed merge, `resource_compute_virtual_machine.go:523-528` / `resource_compute_iaas_opensource_virtual_machine.go:279-284`), so a M1→M2 rename leaves the M1-held registration unexcused and refused — but that is precisely what v1.12.0 already did on xoa/vmware, so it is a pre-existing refusal, not a regression. The remaining attack surface is false PASS only, and I found two narrow vectors, neither unsound: (a) `mac_address` is `Optional`, so a user who *types* a foreign MAC into the block spoofs ownership — but the same apply then attempts a duplicate-MAC PATCH the platform will reject, and it is a deliberate act against oneself; (b) `inlineAdapterMACs` (`:214-226`) flattens all blocks into one unindexed set, so a registration on adapter *j*'s MAC excuses an address configured on adapter *i* — reachable only when two distinct `network_id`s resolve to the same VPC private network (`rejectInlineAdapterIPSharedNetwork`, `:404-425`, already refuses the same-`network_id` case), and it is the identical cross-index semantics the machine-link predicate has always had. MINOR, not blocking. The doc comment's supporting facts check out: the MAC-addressed endpoint is real (`internal/client/vpc_static_ip.go:298`, `GET /vpc/v1/static_ips/mac/%s`) and the `virtualMachine: null` xoa/vmware shape is real (`internal/client/vpc_unit_test.go:213,236,253`).

**A2. Does the MAC excuse let a taken address through `validateInlineAdapterIPPreconditionsCore`?** Walking both flows: a **no-op update** re-reads the address still registered to our own MAC → excused at `:176` → no refusal, which is the whole point and was broken before. A **relocation** X→Y re-keys the lookup on Y (`:169`, `conflictOf(ctx, networkID, ip)` with `ip` from the raw config, `:162`), so the holder examined is Y's holder, not X's; a foreign holder of Y has a foreign MAC and no matching link → refused at `:184`, and `describeStaticIPHolder` (`:253-255`) names the MAC so the user can act. The only way a taken Y slips through is vector (b) of A1 — Y already held by *this same VM's other adapter* on the same private network — where the PATCH would indeed be silently ignored. That is a genuine but narrow widening of an existing hole, and the check is still ordered last behind `refuseBeforeAnyMutation` (`:582`), so a refusal never persists planned values. I do not consider it blocking: it requires an exotic topology and is the same semantic the VM Instances machine-link excuse already carried.

**A3. Anything beyond (i) deferral, (ii) MAC ownership, (iii) hint, (iv) wording?** No. I enumerated every call site: `vpc_inline_adapter_ip.go:371` (plan hook, now passes `diff.Id()` + MACs), `:582` (Compute update gate), and `resource_compute_iaas_opensource_virtual_machine.go:602,684` (the two OpenIaaS creates, `""` + `nil` — behaviour unchanged bar the hint). The `customizeVMInstanceDiff` split (`resource_public_cloud_vm_instance.go:243-289`) is a pure closure extraction: I compared statement-for-statement against the diff hunk — same order, same conditions, same returns — and the production wiring still points at the un-injected wrapper (`:59` → `:246-248`). `createVMInstanceWith`'s refusal (`:524`) gains only `%s` + the hint, whose constant leads with a space so the concatenation reads correctly. One accounting note the brief does not spell out: the hint now also lands on the **VMware and OpenIaaS** create refusals, where `ip_address` is *not* ForceNew — it is still conditionally phrased ("If this registration belongs to…") and a replacement there can still be triggered by another ForceNew attribute, so it is not untrue, just occasionally noise. NIT.

**A4. Are the tests non-vacuous, and is the RED evidence sufficient?** Yes. `recordingConflict` (test file lines 24-35) plus `assertCalls` makes every expected-pass case assert the exact IPAM calls made, so a config that failed to reach the hook cannot pass vacuously — and the deferral cases assert *zero* calls, which is only reachable past `diff.Id() == ""` at `vpc_inline_adapter_ip.go:348`. I re-derived both mutations from the code rather than trusting the brief: removing `:348-354` makes the null-prior cases call `conflictOf` with `ownerVMID == ""`, hit `:181` and refuse → the four named subtests fail, and the identity cases are untouched → the brief's MUTATION 1 ledger is exactly what the code implies. Removing the MAC branch `:199-207` leaves `holder.VirtualMachine == nil` unexcused → precisely the two named subtests fail → MUTATION 2 checks out. On harness fidelity I pushed hard on `proposedNewFor` (test lines 68-84), which approximates `objchange.ProposedNew` at the **top level only** and does not recurse into blocks: the adapter's `mac_address` is therefore null in the shimmed `ResourceConfig`, and the MAC that reaches the hook arrives via the SDK's own Optional+Computed state merge in `SimpleDiff`. That makes the test *stricter* than production (where real `ProposedNew` would also carry it), so the approximation cannot hide a failure. `TestSchemaSnapshot` passing is not evidence about the description text — the snapshot deliberately excludes `Description` (`internal/provider/schema_snapshot_test.go:26`), which is exactly why the drift in A5 survived to here.

**A5. Wording.** The CHANGELOG's factual claims all verify: `cloud_init` is ForceNew (`resource_public_cloud_vm_instance.go:159-162`), `os_network_adapter` is ForceNew (`:126-129`), the date matches today, and moving the `latest` badge to 1.12.1 while leaving 1.12.0 bare matches the file's own convention (1.11.0 through 1.6.0 carry no badge; `stable` sits on 1.5.5). The round-2 NIT is honoured in the CHANGELOG, which says the platform "reclaims the registration shortly after the destroy rather than instantly". **But the brief's claim that `go generate ./...` is idempotent is false.** `resource_public_cloud_vm_instance.go:154` says "(an in-place plan, or the first pass of a replacement plan)" while the generated `docs/resources/public_cloud_vm_instance.md:163` still says "(an in-place plan)" — the docs were regenerated *before* the round-2 NIT fix and never after. `.github/workflows/test.yml:114-118` runs `go generate ./...` then `git diff --exit-code`, so the `generate` job goes red, and the published documentation understates the shipped behaviour. That is BLOCKING (one command). Separately, the hook doc comment at `vpc_inline_adapter_ip.go:326` asserts "(the registration is reclaimed by then)" as a completed fact, contradicting `:232-234` and the CHANGELOG three screens away — the round-2 NIT recurring in a second location. And the CHANGELOG's "the authoritative pre-create check, which is unchanged" is true of the verdict but not of the diagnostic, which now carries the hint.

**A6. State-safety doctrine.** Clean. The change moves in the doctrine's favour, not against it: the only decision that got *more permissive* is a plan-time refusal, which is non-destructive and explicitly licensed by the hook's contract (false PASS allowed, false FAIL forbidden — `:336-339`), and the code that shapes persisted state is untouched. No destructive decision now rests on a weaker signal: the destroy in a destroy-before-create replacement is authorised by Terraform's ForceNew, never by this hook, and the create that follows is still gated by the unchanged pre-create check, which fails closed on an unreadable IPAM plane (`:170-172`) and is wrapped in `refuseBeforeAnyMutation` at `:582`. `withPriorInlineAdapterIPs`, `preserveInlineAdapterIP` and the rollback installation are not in the diff, and `TestInlineAdapterIPRollbackIsInstalledOnEveryEntryPoint` still passes on all four entry points. Nothing here can drop a record.

**A7. Rounds 1-2.** Round 1's unsound `VirtualMachine`-only predicate on xoa/vmware is resolved by the MAC branch, with the fixture evidence I verified independently. Round 2's null-prior false FAIL is resolved at `:348-354` and pinned by the pass-2 and `Resource.Diff` re-diff subtests. Round 2's MINOR (add MAC-owned identity cases) is delivered on both axes — the table cases at `vpc_inline_adapter_ip_collision_test.go:83-118` and the per-surface harness cases. Round 2's first NIT (Description must name the first replacement pass) is done **in the Go source but not propagated to the generated docs** — so it is, in the artefact the user actually reads, still open. That is the same finding as the blocker below. Round 2's second NIT (CHANGELOG must not claim reclamation completed) is honoured in the CHANGELOG but reintroduced in the hook's doc comment.

VERDICT: NO-GO

BLOCKING: docs/resources/public_cloud_vm_instance.md:163 — stale against the schema Description at internal/provider/resource_public_cloud_vm_instance.go:154 (missing "or the first pass of a replacement plan"). `.github/workflows/test.yml:114-118` runs `go generate ./...` and fails on any diff, so the `generate` job is red; the published docs also understate the shipped behaviour. Fix is one command, `go generate ./...`, plus committing the regenerated file. The design itself is sound and nothing else blocks — re-run and this is a GO.

MINOR:
- internal/provider/vpc_inline_adapter_ip.go:214-226 — `inlineAdapterMACs` flattens all blocks into one unindexed set, so a registration on adapter *j*'s MAC excuses an `ip_address` configured on adapter *i*. Reachable only when two distinct `network_id`s resolve to the same VPC private network, and semantically identical to the pre-existing machine-link excuse, but it is a false PASS at the authoritative update gate. Worth a comment saying the excuse is deliberately per-VM, not per-adapter.
- internal/provider/vpc_inline_adapter_ip.go:326 — "(the registration is reclaimed by then)" states as completed what :232-234 and the CHANGELOG both say is not always immediate. The round-2 NIT, recurring. Doc comment only, no behaviour.
- internal/provider/vpc_inline_adapter_ip.go:202-206 — `mac_address` is `Optional` on both Compute surfaces, so `ownerMACs` can carry a user-authored value rather than one read back from the live adapter. The doc comment's "as positive as the machine link" (:145) is therefore slightly stronger than the mechanism warrants; the platform's duplicate-MAC rejection is what actually backstops it, and saying so would make the claim honest.

NIT:
- CHANGELOG.md:8 — "the authoritative pre-create check, which is unchanged" is true of the verdict but not of the diagnostic, which now carries `inlineAdapterIPReplacementHint`.
- internal/provider/vpc_inline_adapter_ip.go:238 — the replacement hint now also reaches the VMware/OpenIaaS create refusals, where `ip_address` is not ForceNew and a replacement is a rarer cause. Conditionally phrased, so not untrue, just occasionally noise.
- internal/provider/vpc_inline_adapter_ip.go:582 — `inlineAdapterMACs(planned)` is computed on the create path too, where `ownerVMID == ""` makes it dead. Harmless, one allocation.

---

## Delta re-review (same reviewer, after the round-3 findings were applied) — verbatim

REVIEWER: Claude Opus 5 (fallback reviewer — delta re-review)

**Chain of custody.** `533_tracked.diff` is byte-identical to the diff I reviewed in round 3, and `533_tracked_v2.diff` is byte-identical to `git diff` on the worktree. `533_delta_since_r3.diff` is therefore an authoritative and complete account of what changed in the tracked files — and I confirmed separately that the untracked `internal/provider/vpc_inline_adapter_ip_replacement_test.go`, which appears in neither diff, is unchanged since round 3. The file set is unchanged (7 modified, 1 untracked). Nothing landed outside the six declared items.

**1. BLOCKING resolved.** I re-ran the exact equality check that found the drift: the `ip_address` Description at `internal/provider/resource_public_cloud_vm_instance.go:154` and the generated line at `docs/resources/public_cloud_vm_instance.md:163` are now byte-for-byte identical, and the docs carry "or the first pass of a replacement plan". I did not run `go generate` myself (read-only brief), so I verified the substance rather than the generator: the whole v2 diff changes exactly **one** `Description:` string and touches exactly **one** docs file, so there is no second drift source for the `generate` job at `.github/workflows/test.yml:114-118` to surface. The shared `inlineAdapterIPDescription` is still untouched, so the two Compute docs pages are unaffected.

**2-5. Documentation-only, and all four now say something true.** `vpc_inline_adapter_ip.go:145-148` drops "as positive as the machine link" for an honest account — the planned `mac_address` is Optional+Computed so it is normally read back from the live adapter, and a typed foreign MAC is backstopped by the platform's duplicate-MAC refusal, not by this check. That is exactly the mechanism I identified and it no longer overclaims. `:150-154` states the per-VM scoping and, better than I asked for, gives the *doctrinal* reason for preferring the laxer predicate: a per-block match could false-FAIL, which the hook's contract forbids outright — so the looser excuse is contract-compliant by construction rather than merely tolerated. `:219-224` mirrors it on `inlineAdapterMACs`. `:337-339` replaces the "(the registration is reclaimed by then)" assertion with "not always instantly — see inlineAdapterIPReplacementHint", removing the contradiction against `:232-234`. `CHANGELOG.md:8` now reads "whose verdict is unchanged (its diagnostic now also says what to do when the holder is the machine being replaced)", which is precisely accurate. No new claim in any of the four is unverifiable or false.

**6. The one code change is a genuine no-op.** `validateInlineAdapterIPPreconditionsCore` now collects `ownerMACs` only when `ownerVMID != ""` (`vpc_inline_adapter_ip.go:594-599`). The callee short-circuits at `:185` on `ownerVMID != "" && staticIPOwnedBy(...)`, so when the owner is empty `ownerMACs` was never read — passing `nil` instead of a populated slice is observationally identical, and `staticIPOwnedBy` ranges over the slice anyway, which is safe on nil. The predicate itself (`:192-208`) is untouched, so my round-3 soundness argument — the new excuse is a strict superset of the v1.12.0 one, hence no new false FAIL is reachable — still holds verbatim. The plan-hook site at `:383` still collects MACs unconditionally, which is correct and not an oversight: it sits behind the `diff.Id() == ""` early return at `:360-366`, so the owner there is never empty and there is no dead input to guard.

**Gates re-run on the current tree.** `gofmt -l internal/` empty; `go vet ./internal/...` clean; the whole package green (`go test ./internal/provider -count=1` → `ok`, 0.713s), and the targeted #533 suite green on its own. Pinned `staticcheck 2025.1.1` under the 1.24 toolchain is clean over `./...`. `TestSchemaSnapshot` still passes — worth restating that this is *not* what protects the description text (the snapshot deliberately excludes `Description`, `schema_snapshot_test.go:26`); the `generate` job is, and it now has nothing to regenerate.

**Residual.** My round-3 MINOR on cross-index (per-VM) scope is not closed by code — by design, and I agree with that call: it is documented, it matches the pre-existing machine-link semantics, and tightening it would trade an allowed false PASS for a forbidden false FAIL. The remaining exposure stays what I described: an address held by this VM's *other* adapter excuses a push, reachable only when two distinct `network_id`s resolve to the same VPC private network. Narrow, pre-existing in kind, and now explicitly a documented choice rather than an accident.

VERDICT: GO

BLOCKING: none

MINOR: none — the three round-3 MINORs are addressed. The per-VM scope of the ownership excuse (`internal/provider/vpc_inline_adapter_ip.go:150-154`, `:219-224`) is now an argued, documented decision rather than an unexamined one; I accept it, and note only that if a per-block match is ever wanted it must come with a way to align planned block order to live adapters, or it will violate the hook's no-false-FAIL contract.

NIT: `internal/provider/vpc_inline_adapter_ip.go:238` — the replacement hint still reaches the VMware/OpenIaaS create refusals, where `ip_address` is not ForceNew. Accepted as is per item 7; the wording is conditional ("If this registration belongs to…") so it is never untrue, and a replacement on those surfaces can still be triggered by another ForceNew attribute.

---

`Awaiting independent reviewer` — the committed diff still owes its pass by the project's declared reviewer (Codex), blocked on the workspace spend cap; the human review of this PR may stand in for it. Merge remains conditioned on `HOTFIX-APPROVED: #533` and a human review.

pbesret:
## Live end-to-end evidence

Reproduction of the issue against the platform, then the fix, with Terraform 1.11.3. Target: one `cloudtemple_public_cloud_vm_instance` (`tf-533-repro`, Debian 13, Development 1 vCPU / 2 GB, `power_state = "off"`) on the VPC private network `pbt-pn-01` (`10.0.5.0/24`, VPC `fsn-01`, zone `fr1-az01`) with the explicit `ip_address = "10.0.5.250"` — the same network as the 25-way concurrent create that motivated the 1.12.0 check. Everything created by the run was destroyed at the end.

| Phase | Provider | Action | Result |
|---|---|---|---|
| B | **1.12.0 (registry)** | `apply`: create the stopped VM with `ip_address` | Created in 2m27s, id `b8012191-…`; IPAM shows `10.0.5.250` registered, source `vmi`, MAC `26:16:cc:fd:ea:e9` |
| C | **1.12.0 (registry)** | `plan` after adding `cloud_init` | **Reproduced**: `Error: os_network_adapter[0] requests ip_address "10.0.5.250", which is ALREADY registered on that VPC private network, registered by vmi for virtual machine b8012191-… on MAC 26:16:cc:fd:ea:e9 …` — the VM's own registration |
| D | **hotfix (this PR, `30e647b`)** | same `plan` (dev_overrides) | **Fixed**: `cloudtemple_public_cloud_vm_instance.repro533 must be replaced`, `cloud_init … forces replacement`, `Plan: 1 to add, 0 to change, 1 to destroy` |
| E | hotfix | `apply` the saved replacement plan (destroy-before-create) | **Converged in one pass**, 151 s end to end: old VM destroyed in 33 s, new VM created in 1m57s (`d57140d4-…`), `Apply complete! Resources: 1 added, 0 changed, 1 destroyed.` The pre-create gate did **not** refuse: by the time the create ran, the destroyed VM's registration had already been reclaimed. IPAM right after: `10.0.5.250` held by the **new** VM (`vmi`). A follow-up `plan` answers `No changes.` |
| F | hotfix | `destroy`, then IPAM sweep | Destroyed in 33 s (`Destroy complete! Resources: 1 destroyed.`). IPAM sweep immediately after: `10.0.5.250` **not registered**; again 45 s later: not registered. No `tf-533*` VM left in the tenant; the local state is empty. Both IPAM and compute planes are clean. |

Additional live observation, relevant to the round-2 review finding: on this DEV tenant the static-IP registrations of source `xoa` and `vmware` **do** carry a `virtual_machine_id` (for example `10.0.6.10 xoa` → VM `291bf0ec…`, `10.0.6.11 vmware` → VM `08982e10…`); only `custom` registrations have none. The `virtualMachine: null` shape carried by the client fixtures was therefore not representative of this environment, and the MAC-ownership proof added here is a second, independent proof rather than the only one that can work.

Harness: the provider was pointed at the API with `api_suffix = false` (the DEV API is served without the `/api` prefix); credentials were sourced from the operator's local secrets file at run time and never written to any file or log.

pbesret:
HOTFIX-APPROVED: #533

Approval given by Paul Besret (authorized human) for this direct hotfix to `main`, in accordance with the RC-flow exception of the engineering rules (blocking client regression on the released v1.12.0, plan-time only, apply-time gate unchanged, live-validated on DEV — see the evidence comment above). Posted on his explicit instruction; the merge and the `v1.12.1` tag are performed by him.

